PR #2, opened 2022-07-21
Babel setup

oluoluoxenfree:
## Steps to test/reproduce

## Show me

### DIFF
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -1,0 +1,6 @@
+module.exports = {
+  presets: [
+    ["@babel/preset-env", { targets: { node: "current" } }],
+    "@babel/preset-typescript",
+  ],
+};
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -1,6 +1,5 @@
 module.exports = {
   presets: [
-    ["@babel/preset-env", { targets: { node: "current" } }],
-    "@babel/preset-typescript",
-  ],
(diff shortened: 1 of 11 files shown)


babel.config.cjs
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   presets: [
-    ["@babel/preset-env", { targets: { node: "current" } }],
-    "@babel/preset-typescript",
```
sanajaved7, line 4: @oluoluoxenfree if we're using typescript, should we still keep `@babel/preset-typescript` here?

oluoluoxenfree: I thought since we're only changing the floating-ui package that it wasn't needed?

sanajaved7: Oh but all of our logic, outside of floating-ui, will be in TS/JS so that would need to be transformed too, right?

oluoluoxenfree: @sanajaved7 i thought that esbuild was handling that, or what does it do? *goes to check*

sanajaved7: btw @oluoluoxenfree want to add that bable/esbuild plugin as part of this branch? I think after that we could go ahead and merge. What do you think?

sanajaved7:
[Refine build process](https://trello.com/c/sTQjtl80/39-refine-build-process)

oluoluoxenfree:
I thought that's what esbuild did?

PR #12, opened 2022-08-15
Simpler type def publishing.

jgerigmeyer:


package.json
```diff
@@ -93,5 +92,8 @@
     "ignore": [
       "@types/node"
     ]
+  },
+  "dependencies": {
+    "@types/css-tree": "^1.0.7"
```
jgerigmeyer, line 97: I'm curious what people think about this. Since we're publishing TS definitions which rely upon `@types/css-tree`, I think this is probably The Right Approach™ -- but it is an unnecessary package for most consumers. 🤷

If we end up exposing any API that relies on types from other dependencies (e.g. `@floating-ui/dom`), we'll want to move those here too.

jgerigmeyer:
@VicGUTT I'd be happy for your review here 👍

VicGUTT:
Hey! Thanks for the ping on this :v:

First things first, that.... fox! 😳

Second things second, looks good to me and I can run the code although I need to do a few tweaks. The fact I'm on Windows and using NPM instead of Yarn are probably the culprit.

The said tweaks are:
- Changing single quotes with escaped double quotes from the `lint:css` script: `"lint:css": "stylelint \"**/*.css\""`
- From what I could gather, Windows seem to have a different mechanism for setting terminal environnement variables, they need to be prefixed by `SET`: `"test:unit": "SET NODE_OPTIONS=--experimental-vm-modules && jest",` 
- In `playwright.config.ts`, most likely due to me using NPM, I need to pass an extra `--` in order for it to boot up the dev server: `command: 'npm run serve -- --port 4000 -l warn',`
- I hade to tweak Jest's test matching pattern: `testMatch: ['<rootDir>/tests/unit/**/.(spec|test).[tj]s?(x)'],` (replaced `?(*.)+(spec|test)` with `.(spec|test)`) in order for it to find the relevant tests.
- `clearMocks: true,` from `jest.config.ts` causes the tests to fail with the following error: `ReferenceError: exports is not defined`. Setting the option to `false` or removing it fixes the problem.

[A commit](https://github.com/oddbird/css-anchor-positioning/commit/85abea0a107cb4d83ef09798ef354a2771014afa) of the changes I had to make.

Note, I'm not really asking for any changes to be made, as the problem seem to be with my setup. If easy fixes/solutions can be found, then great, otherwise it's fine. I'm mostly documenting the issues I found so that others may be aware of and for me to remember later on ^^.

jgerigmeyer:
@VicGUTT Thanks! Can you give this PR another try on Windows?

> The fact I'm on Windows and using NPM instead of Yarn are probably the culprit.

Is it possible to use Yarn on Windows? I'm not tied to Yarn specifically, so we could use NPM if that's more accessible for Windows devs. I went ahead and made that change in https://github.com/oddbird/css-anchor-positioning/pull/12/commits/e06ccf716f8ab7186f0ced7391ee1c5216641142

>     * Changing single quotes with escaped double quotes from the `lint:css` script: `"lint:css": "stylelint \"**/*.css\""`

Fixed.

>     * From what I could gather, Windows seem to have a different mechanism for setting terminal environnement variables, they need to be prefixed by `SET`: `"test:unit": "SET NODE_OPTIONS=--experimental-vm-modules && jest",`

Fixed.

>     * In `playwright.config.ts`, most likely due to me using NPM, I need to pass an extra `--` in order for it to boot up the dev server: `command: 'npm run serve -- --port 4000 -l warn',`

Fixed.

>     * I hade to tweak Jest's test matching pattern: `testMatch: ['<rootDir>/tests/unit/**/.(spec|test).[tj]s?(x)'],` (replaced `?(*.)+(spec|test)` with `.(spec|test)`) in order for it to find the relevant tests.

Interesting. Seems to be a [known issue](https://github.com/facebook/jest/issues/7914#issuecomment-464352069). Hopefully fixed in https://github.com/oddbird/css-anchor-positioning/pull/12/commits/e06ccf716f8ab7186f0ced7391ee1c5216641142

>     * `clearMocks: true,` from `jest.config.ts` causes the tests to fail with the following error: `ReferenceError: exports is not defined`. Setting the option to `false` or removing it fixes the problem.

Interesting -- hopefully fixed in https://github.com/oddbird/css-anchor-positioning/pull/12/commits/e06ccf716f8ab7186f0ced7391ee1c5216641142

> Note, I'm not really asking for any changes to be made, as the problem seem to be with my setup. If easy fixes/solutions can be found, then great, otherwise it's fine. I'm mostly documenting the issues I found so that others may be aware of and for me to remember later on ^^.

Understood, but since we want to use this as a template for future polyfills, I really want this setup to work for many contributors -- thanks for your help testing!

package.json
jgerigmeyer: @oluoluoxenfree @sanajaved7 @jerivas Any thoughts on this question of how to treat TS dependencies?

jgerigmeyer: Anyone using TS who installs this package will not get the correct types for dependencies unless we list them under `dependencies`. But TS is sort of optional, and the type defs aren't strictly required to use our package.

I think the convenience of including type defs outweighs the minimal added size to our installed package (only in dev -- the production scripts are unchanged), but wanted to see what others think.

VicGUTT:
@jgerigmeyer Yep, all good ! Thanks for the changes, specifically wasn't expecting Yarn to be ditched ^^.

Yes Yarn can be used on Windows _(as far as I can tell)_, I've just never had to use it before and therefore never looked into it. NPM is the default and it's the path of least resistance for me ^^.

Good call on using `cross-env`, I totally forgot it's existence.

I made a small [PR here](https://github.com/oddbird/css-anchor-positioning/pull/14) fixing the last remaining issues I saw.

I'm very happy with the changes and they will definitely make my future upcoming contributions easier, so thank you very much for that ! 

---

_Oh Ooh... me continuing to be annoying, look at [what I did](https://github.com/oddbird/css-anchor-positioning/commit/69db15ec59e0af47e52a3a2ce1e5f462027634cb) to have the most perfec**test** test setup 👀_

jgerigmeyer:
@VicGUTT Hopefully final review? 🙏

package.json
```diff
@@ -36,35 +36,35 @@
...
     "format": "run-p format:*",
-    "lint:css": "stylelint '**/*.css'",
+    "lint:css": "stylelint \"**/*.css\"",
     "lint:js": "run-s prettier:check eslint:check tsc",
     "lint": "run-p lint:*",
-    "prepack": "yarn build",
-    "test:unit": "NODE_OPTIONS=--experimental-vm-modules jest",
-    "test:unit:ci": "yarn test:unit --ci",
+    "prepack": "run-s build types",
+    "test:unit": "vitest",
```
jgerigmeyer, line 54: @VicGUTT It looks to me like vitest does not require this `NODE_OPTIONS` setting -- can you confirm that it works on Windows too?

package.json
sanajaved7: +1 to making the type defs available!

sanajaved7 (approved):

package.json
VicGUTT: I can confirm, all good :v:

VicGUTT:
@jgerigmeyer We are good to go ! 🎉 

Thanks for the changes/improvements/openness !

PR #13, opened 2022-08-16
changing fetch to return an array of css strings

oluoluoxenfree:
## Steps to test/reproduce

## Show me


src/fetching.ts
```diff
@@ -1,21 +1,32 @@
...
       CSSlinks.push(link);
     }
   });
-  return CSSlinks;
+
+  const linkedCss = CSSlinks.map((link) => {
+    // fetch css and push into array of strings
+    console.log(link.href);
+    http.get(link.href).toString();
+  });
```
jgerigmeyer, line 19: This `.map` call needs to return something for each item, otherwise it results in e.g. `[undefined]`.

src/fetching.ts
```diff
@@ -1,21 +1,32 @@
...
+    if ((link.type === 'text/css' || link.rel === 'stylesheet') && link.href) {
       CSSlinks.push(link);
     }
   });
-  return CSSlinks;
+
+  const linkedCss = CSSlinks.map((link) => {
+    // fetch css and push into array of strings
+    console.log(link.href);
+    http.get(link.href).toString();
```
jgerigmeyer, line 18: `http.get()` is asynchronous, so to get the string here we'll need to wait for response data, right?

https://nodejs.org/docs/latest-v16.x/api/http.html#httpgeturl-options-callback

index.html
```diff
@@ -21,7 +21,7 @@ <h1>CSS Anchor Positioning Polyfill</h1>
     </popup>
 
     <script type="module">
-      import transformCSS from '/src/index.ts';
+      import transformCSS from './dist/css-anchor-positioning.js';
```
jgerigmeyer, line 24: Why is this change necessary?

index.html
```diff
@@ -5,8 +5,8 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>CSS Anchor Positioning Polyfill Demo</title>
-    <link rel="stylesheet" href="/src/base.css" />
-    <link type="text/css" href="/src/styles.css" />
+    <link rel="stylesheet" href="./src/base.css" />
+    <link type="text/css" href="./src/styles.css" />
```
jgerigmeyer, line 9: Were the absolute URLs not working here?

src/fetching.ts
```diff
@@ -2,31 +2,52 @@ import { computePosition } from '@floating-ui/dom';
...
-  return CSSlinks;
+
+  const linkedCSS = await Promise.all(
+    CSSlinks.map(async (link) => {
+      // fetch css and push into array of strings
+      const response = await fetch(link.href);
+      const text = await response.text();
+      return text;
+    }),
+  );
```
jgerigmeyer, line 21: @oluoluoxenfree When using `Array.prototype.map()` with an asynchronous callback, the array is an array of Promises -- so in this case we want to wrap with `await Promise.all()` to get the underlying resolved values.

https://stackoverflow.com/a/40140562

src/index.ts
```diff
@@ -1,3 +1,6 @@
+// Polyfill `fetch` for older browsers
+import 'whatwg-fetch';
```
jgerigmeyer, line 2: I'm not actually sure if this is needed or not? `fetch` is supported by all ES6 browsers AFAIK, and that's our current support target.

index.html
```diff
@@ -5,8 +5,13 @@
...
     <title>CSS Anchor Positioning Polyfill Demo</title>
-    <link rel="stylesheet" href="/src/base.css" />
-    <link type="text/css" href="/src/styles.css" />
+    <link rel="stylesheet" href="/base.css" />
+    <link type="text/css" href="/styles.css" />
+    <style>
+      .test {
+        color: red;
+      }
+    </style>
```
jgerigmeyer, line 14: @oluoluoxenfree Feel free to remove/change this -- I just wanted to test that this was functionally working.

index.html
```diff
@@ -5,8 +5,13 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>CSS Anchor Positioning Polyfill Demo</title>
-    <link rel="stylesheet" href="/src/base.css" />
-    <link type="text/css" href="/src/styles.css" />
+    <link rel="stylesheet" href="/base.css" />
+    <link type="text/css" href="/styles.css" />
```
jgerigmeyer, line 9: @oluoluoxenfree The main fix was moving these styles into the `public/` dir. By default, Vite does a lot of fancy pre-processing and auto-injection when CSS files are imported or fetched. By moving them to `public/` we can tell Vite to serve these assets raw, which is I think what we want for this polyfill.

See:
- https://vitejs.dev/guide/features.html#css
- https://vitejs.dev/guide/assets.html#the-public-directory

src/fetching.ts
jgerigmeyer: It strikes me that most of these CSS polyfills would be doing something similar -- finding all the styles in the document, and extracting them as strings. Is there code in the other existing polyfills we could copy?

package.json
```diff
@@ -91,7 +93,8 @@
   },
   "yarn-upgrade-all": {
     "ignore": [
-      "@types/node"
+      "@types/node",
+      "node-fetch"
```
sanajaved7, line 97: I think you said we didn't need node-fetch after all right?

src/fetching.ts
```diff
@@ -2,10 +2,14 @@ import { computePosition } from '@floating-ui/dom';
 
 import type { PositionFallbackRulesMap } from './parsing.js';
 
+export const CSSlinks: HTMLLinkElement[] = [];
```
jgerigmeyer, line 5: @oluoluoxenfree I'm curious why this array is moved top-level and exported?

package.json
jgerigmeyer: This is required to run the tests -- what we didn't need (and Olu already removed) was the production whatwg-fetch polyfill.

src/fetching.ts
oluoluoxenfree: @sanajaved was saying preserving the order and destination of the links was important to what she was doing, so I was making sure it's available. is there an easier way to do that?

jgerigmeyer (approved):

src/fetching.ts
sanajaved7: +1 to Olu and for some more context - when transforming the original styles in a stylesheet - it would helpful to know the href of the stylesheet that particular CSS string came from since after transformation, I need to replace that href URL with the new source URL. I could probably rework this for what I need though if we want to go ahead and merge this as-is. But I was hoping we wouldn't need to write two separate CSS fetching functions.

jgerigmeyer: Ah, ok. What about including the src href in the returned list? Something like https://github.com/oddbird/css-anchor-positioning/pull/13/commits/9dd04517ef991db1a07f43b5bd4afe7b23117729

sanajaved7: That's perfect - thanks! 🎉

sanajaved7:
[Make fetchCSS command return (an array/object of) strings](https://trello.com/c/gbU3qRHx/43-make-fetchcss-command-return-an-array-object-of-strings)

sanajaved7 (approved):

jgerigmeyer (approved):

PR #14, opened 2022-08-17
fix: passing arguments to npm-run-script

VicGUTT:
## Steps to test/reproduce
Without the `--` _(double dashes)_ the flags passed to the NPM run scripts from another script are not interpreted.

From the [NPM docs](https://docs.npmjs.com/cli/v8/commands/npm-run-script):

>  Use `--` to pass `-` -prefixed flags and options which would otherwise be parsed by npm.

## Show me
Nothing to show :)


jgerigmeyer:
Incorporated these changes into #12

PR #16, opened 2022-08-19
Flesh out sample CSS

jgerigmeyer:
The goal here is to define the sample CSS for each use-case our polyfill is trying to address. Ideally we would have separate elements on the demo page for each of these, or a toggle that switches in between -- but that's maybe not urgent.

@sanajaved7 @oluoluoxenfree I think the code here is enough to write unit tests, but the CSS isn't fleshed out to actually make these functional demos. Feel free to add that in now, or we can do it later.


jgerigmeyer:
These examples are really helpful: https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/web_tests/wpt_internal/css/css-anchor-position

I'm going to update the ones here to match those...

jgerigmeyer (commented):
I want our sample CSS to be relatively robust (for unit testing), but I don't think our demo page or e2e tests need to cover every edge case. Ideally we want to use the existing WPTs for that.

sanajaved7 (approved):

oluoluoxenfree (approved):

public/position-fallback.css
```diff
@@ -0,0 +1,50 @@
...
+    left: anchor(--my-anchor-fallback left);
+    bottom: anchor(--my-anchor-fallback top);
+  }
+
+  @try {
+    /* 5: Position to the left with the narrower width. */
+    left: anchor(--my-anchor-fallback right);
+    top: anchor(--my-anchor-fallback top);
+    width: 35px;
+    height: 40px;
```
jgerigmeyer, line 48: @sanajaved7 @oluoluoxenfree I realized that we'll need to pay attention to this when we do the transform (or during the code-removal) -- we want to remove the positioning declarations, but any other arbitrary declarations in the relevant `@try` block will probably need to be moved up into the end-result stylesheet.

PR #18, opened 2022-08-22
Re-work CSS parsing logic.

jgerigmeyer:

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -1,63 +1,156 @@
 import * as csstree from 'css-tree';
 
-interface TryBlockMap {
-  [key: string]: string;
+interface DeclarationWithValue extends csstree.Declaration {
+  value: csstree.Value;
 }
 
-interface PositionFallbackRulesMap {
-  [key: string]: TryBlockMap[];
+interface AtRuleRaw extends csstree.Atrule {
+  prelude: csstree.Raw | null;
 }
 
-export function isFallbackDeclaration(node: csstree.CssNode) {
-  return (
-    node.type === 'Declaration' && node.property.includes('position-fallback')
-  );
+export function isDeclaration(
+  node: csstree.CssNode,
+): node is DeclarationWithValue {
+  return node.type === 'Declaration';
 }
 
-export function isFallbackAtRule(node: csstree.CssNode) {
-  return node.type === 'Atrule' && node.name.includes('position-fallback');
+export function isAnchorNameDeclaration(
+  node: csstree.CssNode,
+): node is DeclarationWithValue {
+  return node.type === 'Declaration' && node.property === 'anchor-name';
 }
 
-function parsePositionFallback(ast: csstree.CssNode) {
-  const parsedFallbackRules = csstree.findAll(
-    ast,
-    isFallbackAtRule,
-  ) as csstree.Atrule[];
-  const fallbacks: PositionFallbackRulesMap = {};
+export function isAnchorFunction(
+  node: csstree.CssNode | null,
+): node is csstree.FunctionNode {
+  return Boolean(node && node.type === 'Function' && node.name === 'anchor');
+}
 
-  if (parsedFallbackRules) {
-    parsedFallbackRules.forEach((fallback) => {
-      const fallbackName = (fallback.prelude as csstree.Raw).value;
-      const fallbackTryBlocks: TryBlockMap[] = [];
+export function isFallbackDeclaration(
+  node: csstree.CssNode,
+): node is DeclarationWithValue {
+  return node.type === 'Declaration' && node.property === 'position-fallback';
+}
 
-      const children = (fallback.block?.children || []) as csstree.Atrule[];
-      children.forEach((childBlock) => {
-        const tryBlock: TryBlockMap = {};
+export function isFallbackAtRule(node: csstree.CssNode): node is AtRuleRaw {
+  return node.type === 'Atrule' && node.name === 'position-fallback';
+}
 
-        if (childBlock.block?.children) {
-          (
-            childBlock.block?.children as csstree.List<csstree.Declaration>
-          ).forEach((child) => {
-            tryBlock[child.property] = (child.value as csstree.Raw).value;
-          });
-        }
+export function isTryAtRule(node: csstree.CssNode): node is AtRuleRaw {
+  return node.type === 'Atrule' && node.name === 'try';
+}
 
-        fallbackTryBlocks.push(tryBlock);
+function parseAnchorFn(node: csstree.FunctionNode) {
+  let anchorName: string | undefined,
+    anchorEdge: string | undefined,
+    fallbackValue = '',
+    foundComma = false;
+  node.children.toArray().forEach((child, idx) => {
+    if (foundComma) {
+      fallbackValue = `${fallbackValue}${csstree.generate(child)}`;
+      return;
+    }
+    if (child.type === 'Operator' && child.value === ',') {
+      foundComma = true;
+      return;
+    }
+    if (child.type === 'Identifier' && child.name) {
+      switch (idx) {
+        case 0:
+          anchorName = child.name;
+          break;
+        case 1:
+          anchorEdge = child.name;
+          break;
+      }
+    }
+  });
+  return {
+    anchorName,
+    anchorEdge,
+    fallbackValue,
+  };
+}
+
+export function getDataFromCSS(css: string) {
+  const anchorNames: any = {};
+  const anchorFunctions: any = [];
+  const fallbackDeclarations: any = {};
+  const fallbacks: any = {};
+  const ast = parseCSS(css);
+  csstree.walk(ast, function (node) {
+    const rule = this.rule?.prelude as csstree.Raw | undefined;
+    if (
+      isAnchorNameDeclaration(node) &&
+      node.value.children.first &&
+      rule?.value
+    ) {
+      const name = (node.value.children.first as unknown as csstree.Identifier)
+        .name;
+      if (anchorNames[name]) {
+        anchorNames[name].push(rule.value);
+      } else {
+        anchorNames[name] = [rule.value];
+      }
+    }
+    if (isAnchorFunction(node) && rule?.value && this.declaration) {
+      const { anchorName, anchorEdge, fallbackValue } = parseAnchorFn(node);
+      anchorFunctions.push({
+        floatingEl: rule.value,
+        floatingEdge: this.declaration.property,
+        anchorName,
+        anchorEdge,
+        fallbackValue,
       });
-      fallbacks[fallbackName] = fallbackTryBlocks;
-    });
-  }
+    }
+    if (
+      isFallbackDeclaration(node) &&
+      node.value.children.first &&
+      rule?.value
+    ) {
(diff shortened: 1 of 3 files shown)


jgerigmeyer:
[Update parsing to start with anchor()](https://trello.com/c/w3XAIWjb/44-update-parsing-to-start-with-anchor)

src/parse.ts
```diff
@@ -1,63 +1,156 @@
...
+        anchorNames[name].push(rule.value);
+      } else {
+        anchorNames[name] = [rule.value];
+      }
+    }
+    if (isAnchorFunction(node) && rule?.value && this.declaration) {
+      const { anchorName, anchorEdge, fallbackValue } = parseAnchorFn(node);
+      anchorFunctions.push({
+        floatingEl: rule.value,
+        floatingEdge: this.declaration.property,
```
jgerigmeyer, line 100: We should check whether this is position-related -- if not, then we need a regular math polyfill for it (instead of positioning).

src/parse.ts
```diff
@@ -1,63 +1,156 @@
...
+      switch (idx) {
+        case 0:
+          anchorName = child.name;
+          break;
+        case 1:
+          anchorEdge = child.name;
+          break;
+      }
+    }
+  });
```
jgerigmeyer, line 67: I don't love this, but I'm not sure the best approach. It's nice to let csstree handle the parsing here, but then we want some of the children back as a plain string. Options as I see it:

1. Turn off value parsing (https://github.com/csstree/csstree/blob/master/docs/parsing.md#parsevalue), in which case we need to detect an anchor-fn with our own logic (e.g. `startsWith('anchor(')` etc). I don't think that's great, since csstree is built to do that for us.
2. Turn on value parsing, but then it splits the optional `<length-percentage>` arg into an arbitary number of child nodes, when really we just want it as a raw string.

I went with option 2 for now, and manually converted it back to a raw string.

src/parse.ts
```diff
@@ -1,63 +1,156 @@
...
+          const tryBlock: any = {};
+          const declarations = atRule.block.children.filter(isDeclaration);
+          declarations.forEach((child) => {
+            const firstChild = child.value.children
+              .first as unknown as csstree.CssNode;
+            if (firstChild && isAnchorFunction(firstChild)) {
+              tryBlock[child.property] = parseAnchorFn(firstChild);
+            } else {
+              tryBlock[child.property] = csstree.generate(child.value);
+            }
```
jgerigmeyer, line 134: I'm not actually sure what format we want for the rules inside these `@try` blocks now -- we should discuss.

src/parse.ts
```diff
@@ -1,63 +1,156 @@
...
+          tryBlocks.push(tryBlock);
+        }
+      });
+      fallbacks[name] = tryBlocks;
+    }
+  });
+  console.log(anchorNames);
+  console.log(anchorFunctions);
+  console.log(fallbackDeclarations);
+  console.log(fallbacks);
```
jgerigmeyer, line 146: We'll probably want to do some sanity-checking (and probably combining) of these various types.

src/parse.ts
```diff
@@ -1,63 +1,156 @@
...
+  const anchorFunctions: any = [];
+  const fallbackDeclarations: any = {};
+  const fallbacks: any = {};
+  const ast = parseCSS(css);
+  csstree.walk(ast, function (node) {
+    const rule = this.rule?.prelude as csstree.Raw | undefined;
+    if (
+      isAnchorNameDeclaration(node) &&
+      node.value.children.first &&
+      rule?.value
```
jgerigmeyer, line 86: I'm not 100% sure if using the `value` of the nearest `rule` is robust enough here. Do we have to care about nested selectors? (Does csstree even support nested selectors?)

jgerigmeyer (commented):
@sanajaved7 @oluoluoxenfree I have a number of questions, and it may make sense to talk through them together on a call -- perhaps tomorrow after check-in?

src/parse.ts
```diff
@@ -1,63 +1,156 @@
...
+}
+
+export function getDataFromCSS(css: string) {
+  const anchorNames: any = {};
+  const anchorFunctions: any = [];
+  const fallbackDeclarations: any = {};
+  const fallbacks: any = {};
+  const ast = parseCSS(css);
+  csstree.walk(ast, function (node) {
+    const rule = this.rule?.prelude as csstree.Raw | undefined;
```
sanajaved7, line 82: I wonder if we could leverage the `find` and `findAll` functions csstree comes with instead of checking each rule ourselves, not sure if you tried that already @jgerigmeyer and encountered some limitations. I'm wondering if we could do something like this for the data we are looking for and then do some kind of combination like you mentioned once we have everything: 
```
export function findAnchorNames(ast) {
  const anchorNames = [];
  const nodes = csstree.findAll(ast, (node) => isAnchorNameDeclaration(node));

  nodes.forEach((node) => anchorNames.push(node.value.value));
  return anchorNames;
}
```

sanajaved7: And maybe `getDataFromCSS` could look something like this? If this could work for our needs, it would be easier to test and read too.  
```
export function getDataFromCSS(css: string) {
 
  const fallbackDeclarations: any = {};
  const fallbacks: any = {};
  const ast = parseCSS(css);

  const anchorFunctions = findAnchorFunctions(ast);
  const anchorNames = findAnchorNames(ast);
  ```

Comment (jgerigmeyer) [reply src/parse.ts:82]: I like that in theory, but the problem is we often need the parent rule/declaration also, and find/findAll don't provide that context. For example, we can use those to get the list of `anchor-name` declarations, but we won't have access to the parent selector to attach to the name itself.

Comment (sanajaved7) [reply src/parse.ts:82]: Yea I was wondering if we could use find / findAll to also get those parent selectors and then do a merge at the end but maybe that's not doable or preferred 🤔

Comment (jgerigmeyer) [reply src/parse.ts:82]: I guess I'm not sure what that would look like. When we find a use of e.g. `anchor()`, how do we then do another find to get the parent of that declaration? That is, what do we use as the basis to find it?

My current understanding is that the only way to get any parent/ancestor info from a node is to use `walk`.

**Comment by jgerigmeyer** [comment]
Putting this on hold for now, while we reconsider our approach to parsing stylesheets (see: https://oddbird.slack.com/archives/C03K136KX2A/p1661273459265879).

File: src/parse.ts [hunk 7]
```diff
@@ -1,63 +1,264 @@
...
+      if (anchorNames[name]) {
+        anchorNames[name].push(rule.value);
+      } else {
+        anchorNames[name] = [rule.value];
+      }
+    }
+    // Parse `anchor()` function
+    if (isAnchorFunction(node) && rule?.value && this.declaration) {
+      const { anchorName, anchorEdge, fallbackValue } = parseAnchorFn(node);
+      if (anchorName && anchorEdge) {
```
jgerigmeyer, line 156: When an `anchor()` fn is invalid, instead of ignoring it we should use the fallback value (or `0px`).

src/parse.ts
```diff
@@ -1,63 +1,327 @@
...
+    // Parse `position-fallback` declaration
+    const { name: fbName, selector: fbSelector } =
+      getPositionFallbackDeclaration(node, rule);
+    if (fbName && fbSelector) {
+      // This will override earlier `position-fallback` declarations
+      // with the same rule selector:
+      // (e.g. multiple `position-fallback:` declarations
+      // for the same `.foo {...}` selector)
+      fallbackNames[fbSelector] = fbName;
+    }
```
jgerigmeyer, line 241: @mirisuzanne I wanted to confirm that this override logic (outlined in the two comments here) is correct for CSS, right? That is, in all of these cases the last one wins, correct? (Don't start about layers and `!important`, I know I'm simplifying for this initial version...)

```css
.foo {
  top: anchor(--my-anchor-positioning bottom, 50px);
  position-fallback: --fallback1;
  position-fallback: --fallback2;
}

.foo {
  top: 10px;
  position-fallback: --fallback3;
}
```

Ends up with:

```css
.foo {
  top: 10px;
  position-fallback: --fallback3;
}
```

src/transform.ts
```diff
@@ -24,21 +29,31 @@ export function removeAnchorCSS(originalCSS: string) {
...
   });
 
+  // Handle inline stylesheets
   const styleTagCSS = document.querySelectorAll('style');
   styleTagCSS.forEach((element) => {
     element.innerHTML = removeAnchorCSS(element.innerHTML);
   });
+
+  // Get data from concatenated styles
+  getDataFromCSS(allCSS.join('\n'));
```
jgerigmeyer, line 58: @sanajaved7 What do you think of this change to parse all the CSS at once?

Changes in: https://github.com/oddbird/css-anchor-positioning/pull/18/commits/5362d07b9c53d1c5058b811dec65bbdfe15447ac

src/parse.ts
```diff
@@ -1,63 +1,327 @@
...
+      // with the same name:
+      // (e.g. multiple `@position-fallback --my-fallback {...}` uses
+      // with the same `--my-fallback` name)
+      fallbacks[fbRuleName] = fbTryBlocks;
+    }
+  });
+
+  // Merge data together under floating-element selector key
+  const validPositions: AnchorPositions = {};
+  for (const [floatingEl, anchorFns] of Object.entries(anchorFunctions)) {
```
jgerigmeyer, line 257: TODO: This currently does not account for floating elements with `position-fallback` but no `anchor()` fns.

src/parse.ts
mirisuzanne: That's right, since the selectors are identical. Harder to determine if two non-identical selectors apply to the same element - but I assume that problem is also left for solving later.

src/fetch.ts
```diff
@@ -9,43 +9,44 @@ export function isStyleLink(link: HTMLLinkElement) {
...
+    if (el.tagName.toLowerCase() === 'link') {
+      const url = getStylesheetUrl(el as HTMLLinkElement);
+      if (url) {
+        sources.push(url);
+      }
+    }
+    if (el.tagName.toLowerCase() === 'style') {
+      sources.push(el.innerHTML);
+    }
+  });
```
jgerigmeyer, line 49: I combined these into one loop so that we can know that we are storing them in source order, and can assume (for now) that later ones override earlier ones in cases of conflict/duplication.

mirisuzanne:
@jgerigmeyer what's most helpful for me to understand or review here? Is a functional review helpful, or is there something I should be digging into code-wise?

jgerigmeyer:
> @jgerigmeyer what's most helpful for me to understand or review here? Is a functional review helpful, or is there something I should be digging into code-wise?

@mirisuzanne There's not a lot functional to review yet (other than looking at what's logged to the console, and comparing that with the demo stylesheets). You could skim the code (mostly the changes in `parse.ts` and `fetch.ts`), and/or take a look at the console logged data object (which should represent an extraction of the relevant data from the stylesheets).

jgerigmeyer:
@sanajaved7 @mirisuzanne I think this is ready for review, then merge into #19?

sanajaved7 (approved):

oluoluoxenfree (approved):

mirisuzanne (approved):
looking good!

PR #19, opened 2022-08-25
Transform for floating UI

oluoluoxenfree:
## Steps to test/reproduce

## Show me


src/transform.ts
```diff
@@ -42,3 +43,113 @@ export async function transformCSS() {
...
+
+    if (anchor && floating) {
+      autoUpdate(anchor, floating, () => {
+        computePosition(anchor, floating, {
+          // @@@ We should convert `anchorEdge` to valid placement options
+          //
+          // @@@ This is way too "smart" -- we're ignoring the property
+          // declaration entirely, and just assuming that the property is the
+          // opposite of the `anchorEdge`. What if we want the `top` of the
+          // floating element to line up with the `top` of the anchor element?
```
jgerigmeyer, line 131: @oluoluoxenfree @sanajaved7 I think this is the biggest outstanding question for using floating-ui -- how do we translate this into something that works with floating-ui?

```css
#floating-el {
  position: absolute;
  top: anchor(--my-anchor top);
  left: anchor(--my-anchor left);
}
```

jgerigmeyer: I guess there are two questions wrapped up into one there:

1. How do we handle situations where the floating edge (`top`) is not opposite from the anchor edge (`top`)?
2. How do we handle multiple position declarations like this, on the same floating element?

src/transform.ts
```diff
@@ -42,3 +43,113 @@ export async function transformCSS() {
...
+    // @@@ How do we handle multiple declarations?
+    const anchorObj = Object.values(value.declarations)[0];
+    const floating: HTMLElement | null = document.querySelector(key);
+    // @@@ For now, assume the first element is valid
+    const anchor = document.querySelector(anchorObj.anchorEl[0]);
+
+    if (anchor && floating) {
+      autoUpdate(anchor, floating, () => {
+        computePosition(anchor, floating, {
+          // @@@ We should convert `anchorEdge` to valid placement options
```
jgerigmeyer, line 126: - https://tabatkins.github.io/specs/css-anchor-position/#typedef-anchor-side
- https://floating-ui.com/docs/computePosition#placement

src/transform.ts
jgerigmeyer: Maybe using `offset` middleware?

https://oddbird.slack.com/archives/C03K136KX2A/p1661454529540349?thread_ts=1661453174.692759&cid=C03K136KX2A

sanajaved7: @jgerigmeyer @oluoluoxenfree some good news: floating-ui provides additional [placements](https://floating-ui.com/docs/computePosition#placement) like `top-end` and `top-start` that for our needs, map nicely to multiple declarations. 

So for example, we would use the `top-start` placement when we have declarations for `top` and `left`. I don't think we even need the offset middleware to use those placements. It would look something like this: 

```
computePosition(reference, floating, {
  // Try changing this to a different side.
  placement: "top-end"
}).then(({ x, y }) => {
  Object.assign(floating.style, {
    top: `${y}px`,
    left: `${x}px`
  });
});
```

Screen Shot 2022-08-26 at 3 26 47 PM

Here's a [codesandbox](https://codesandbox.io/s/busy-fast-v49qug?file=/src/index.js:0-398) if you want to play around with it too. 

If this sounds good to you both, we will need to add a step that checks for multiple declarations and then does some mapping from our declarations to the correct floating-ui placement. From the docs it looks like there are 12 placement options: 

```
type Placement =
  | 'top'
  | 'top-start'
  | 'top-end'
  | 'right'
  | 'right-start'
  | 'right-end'
  | 'bottom'
  | 'bottom-start'
  | 'bottom-end'
  | 'left'
  | 'left-start'
  | 'left-end';
  ```

Comment (jgerigmeyer) [reply src/transform.ts:131]: @sanajaved7 Those are the ones I linked to yesterday, and I think they're only helpful for e.g. `bottom: anchor(--foo top)` + `right: anchor(--foo right)`. I don't see how the existing placements solve `bottom: anchor(--foo top)` + `left: anchor(--foo right)` -- do you? That is, they don't let us do any placements where the floating element overlaps (unless we use negative offset), or where corners touch.

Comment (jgerigmeyer) [reply src/transform.ts:131]: (I still think your conclusion is correct, that we'll need a mapping step. But `-start` and `-end` only work to line up the same edge of both floating and target.)

Comment (sanajaved7) [reply src/transform.ts:131]: Oops sorry for missing your comment re: placement - I think I was focusing on answering the question from yesterday of whether floating would support multiple placements at all. I do think for those (literal 😄) corner cases, we might need to use offset. I can look into that more. Also, I'm curious about the overlapping usecase, do you have WPT or other examples you could point me to? 🙏🏼

Comment (jgerigmeyer) [reply src/transform.ts:131]: > Also, I'm curious about the overlapping usecase, do you have WPT or other examples you could point me to? 🙏🏼

@sanajaved7 I haven't looked through all the WPTs, but I think the example in my first comment here is a simple overlapping case that we'd need to support:

```css
#floating-el {
  position: absolute;
  top: anchor(--my-anchor top);
  left: anchor(--my-anchor left);
}
```

I don't think floating-ui helps us much with _either_ of those declarations individually, unless we use negative offset.

Here's an overly-verbose list/mapping of all the simple cases we should be able to handle, in `floatingEdge / anchorEdge -- proposed floating-ui solution` format:

- top / top -- `top` + vertical `offset`?
- top / bottom -- `bottom`
- bottom / bottom -- `bottom` + vertical `offset`?
- bottom / top -- `top`
- left / left -- `left` + horizontal `offset`?
- left / right -- `right`
- right / right -- `right` + horizontal `offset`?
- right / left -- `left`

I suppose weird mixes like `top / left` are technically valid, but I don't know that they're meaningful or something we need to worry about initially? Similarly when both `top` and `bottom` are declared on the same element. @mirisuzanne Can you confirm or weigh in?

Aside: We also have the issue where `top` is declared as an anchor, but maybe `left` or `right` is declared as plain CSS (`50px`) -- I'm not sure we can even use floating-ui in that case, since it takes over _all_ positioning of the floating element.

But then we also should be able to handle any *combination* of the above:

- top / top + left / left -- `top-start` + vertical offset?
- top / top + left / right -- `right-start`
- top / top + right / right -- `top-end` + vertical offset?
- top / top + right / left -- `left-start`
- top / bottom + left / left -- `bottom-start`
- top / bottom + left / right -- `bottom-end` + horizontal offset? or `right-end` + vertical offset?
- top / bottom + right / right -- `bottom-end`
- top / bottom + right / left -- `bottom-start` + horizontal offset? or `left-end` + vertical offset?

...you get the idea. I think we could look for patterns in that, so we don't have to map out every case.

(I remain slightly skeptical that floating-ui is the best approach long-term, if we want the polyfill to handle various cases. But I say that without knowing how difficult it would be for us to do similar position-calculations on our own 😅)

sanajaved7: FWIW, I think I got the logic for the corner anchoring working.

https://codesandbox.io/s/pedantic-james-75vm18?file=/src/index.js
Screen Shot 2022-08-26 at 5 08 31 PM

sanajaved7: Mostly through combinations of placement + crossAxis shifts like this:  
```
computePosition(reference, floating, {
  // Try changing this to a different side.
  middleware: [
    offset({
      crossAxis: floating.getBoundingClientRect().width
    })
  ],
  placement: "top-end"
}).then(({ x, y }) => {
  Object.assign(floating.style, {
    top: `${y}px`,
    left: `${x}px`
  });
});```

Comment (jgerigmeyer) [reply src/transform.ts:131]: Nice! And I think we can use their calculations in the `offset` middleware to shorten it a bit (see: https://floating-ui.com/docs/offset#function):

```js
computePosition(reference, floating, {
  middleware: [
    offset(({ rects }) => ({ crossAxis: rects.floating.width }))
  ],
  placement: "top-end"
}).then(({ x, y }) => {
  Object.assign(floating.style, {
    top: `${y}px`,
    left: `${x}px`
  });
});
```

jgerigmeyer: @sanajaved7 @oluoluoxenfree  I withdraw some of my skepticism. 😄 

Ok and wait, so here's another not-very-thought-through brainstorm, but I think I've been overly simplistic about how floating-ui works. Couldn't we use floating-ui's `computePosition` to get the values we need for every single `anchor()` call (even ones used for math, not positioning) -- perhaps calling it multiple times -- but then _not_ (or at least not always) do a direct `x, y` positioning with the values it returns? Instead we'd update the CSS with the returned values, or add styles to the relevant element directly, or something like that.

For example, my earlier case:

```css
#floating-el {
  position: absolute;
  top: anchor(--my-anchor top);
  left: anchor(--my-anchor left);
}
```

Instead of trying to handle this by letting floating-ui do the positioning with one call, we could do something like:

```js
Promise.all([
  computePosition(reference, floating, {
    placement: "top-start"
  }),
  computePosition(reference, floating, {
    placement: "left-start"
  })
]).then(([{ x }, { y }]) => {
  // Or we could set these values back to the CSS,
  // and let that do the positioning.
  Object.assign(floating.style, {
    top: `${y}px`,
    left: `${x}px`
  });
});
```

See codesandbox: https://codesandbox.io/s/inspiring-babycat-bwd7w7

In this particular case I'm not sure it has much advantage over the `offset` approach -- but it seems more flexible for cases where we need a math value, but don't necessarily want floating-ui to do the positioning?

sanajaved7: @jgerigmeyer @oluoluoxenfree that would be awesome if we could use `computePosition` for when `anchor()` is used in math / other calculation functions. Have we determined if we always need to provide it with a floating element? I'm wondering what element we would pass in then if we just wanted to use `anchor()` for math rather than positioning? Would it be whatever element the math / calculation is being called on?

sanajaved7: Answering a bit of my own question here 😆 but something like this lets us use the anchor element's width to style a button's width. And I'm passing in the button element where we would normally pass in the floating element.
```
computePosition(reference, button, {
  middleware: [
    offset(({ rects }) => {
      Object.assign(button.style, {
        width: `${rects.reference.width}px`
      });
    })
  ]
});
```

Screen Shot 2022-08-29 at 5 19 59 PM

And I think if we received a percentage like in the spec example (i.e. `--center: anchor(--reference 50%);`) that would be fairly easy to pass in from parsing. 

https://codesandbox.io/s/late-sun-0rdvnl?file=/src/index.js

mirisuzanne: > I suppose weird mixes like top / left are technically valid, but I don't know that they're meaningful or something we need to worry about initially? 

Yeah, those are possible in CSS, but pretty unlikely in practice. Not the highest priority.

> Similarly when both top and bottom are declared on the same element. @mirisuzanne Can you confirm or weigh in?

The spec for absolute positioning is here: https://drafts.csswg.org/css-position/#abspos-layout

Conceptually, the inset properties define a new ‘positioning box’ relative to the original box. And then the floating content is laid out in (and potentially overflows) that new box. And, by default, items stretch to fit the box (see the automatic sizing section after what I linked). If top and bottom are explicit, then the floating element will try to stretch and match both those offset edges. If it’s too big to fit, it will overflow in the direction of text.

But ideally we don't have to overthink that - we resolve to a px value, and let CSS do the positioning.

src/transform.ts
```diff
@@ -55,5 +57,99 @@ export async function transformCSS() {
...
+  },
+  'right:right': {
+    placement: 'top-end',
+    coord: 'x',
+  },
+  'right:left': {
+    placement: 'left',
+    coord: 'x',
+  },
+};
```
jgerigmeyer, line 104: This seems to work well for the basic simple use-cases. When we get to math or custom vars, though, we'll want to replace this with a more agnostic mapping that can return a value for the anchor-edge without knowing anything about the floating-edge. Maybe something like @sanajaved7's example here: https://codesandbox.io/s/late-sun-0rdvnl?file=/src/index.js

jgerigmeyer:
[transformCSS() - convert positioning to JS](https://trello.com/c/2BVeHDpR/37-transformcss-convert-positioning-to-js)

src/transform.ts
jgerigmeyer: Addressed in #22

jgerigmeyer (approved):
@sanajaved7 @oluoluoxenfree I added unit test coverage for the main logic. We can add more e2e tests, but I think it doesn't need to hold up merge -- so I'd say this is good to go once the two of you have reviewed it!

src/polyfill.ts
```diff
@@ -1,12 +1,198 @@
...
+                          anchorRect: rects.reference,
+                          anchorEdge: anchorValue.anchorEdge,
+                          floatingPosition: property,
+                          fallback: anchorValue.fallbackValue,
+                        }),
+                      });
+                      return 0;
+                    }),
+                  ],
+                });
```
jgerigmeyer, line 156: I looked into replace `computePosition` and `offset` with something simpler, but the relevant code (to [get the anchor rectangle](https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/platform.ts#L19-L23)) is not simple, and unfortunately floating-ui doesn't expose it so we can't use just that piece directly.

sanajaved7 (approved):

PR #22, opened 2022-09-07
First attempt transforming directly to pixel values.

jgerigmeyer:
This is an attempt to resolve `anchor()` fns directly to a pixel value, without reference to the floating-el property being declared. Ideally this will allow us to handle math/custom-css-var use-cases (e.g. `--center: anchor(--my-anchor 50%);`), `anchor-size`, etc.

See:

- https://github.com/oddbird/css-anchor-positioning/pull/19#discussion_r965070310
- https://github.com/oddbird/css-anchor-positioning/pull/19#discussion_r957481575

### DIFF
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -1,4 +1,11 @@
-import { autoUpdate, Axis, computePosition, Placement } from '@floating-ui/dom';
+import {
+  autoUpdate,
+  Axis,
+  computePosition,
+  offset,
+  Placement,
+  Rect,
+} from '@floating-ui/dom';
 import * as csstree from 'css-tree';
 
 import { fetchCSS, isStyleLink } from './fetch.js';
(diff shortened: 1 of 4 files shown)


src/transform.ts
```diff
@@ -63,44 +65,115 @@ export async function transformCSS() {
...
+  switch (position) {
+    case 'top':
+    case 'bottom':
+      return 'y';
+    case 'left':
+    case 'right':
+      return 'x';
+  }
+  return null;
+};
```
jgerigmeyer, line 103: @mirisuzanne Is this reliable? Or is there a better way to determine the inline/block axis?

src/transform.ts
```diff
@@ -63,44 +65,115 @@ export async function transformCSS() {
...
+  anchorRect: Rect;
+  anchorEdge?: AnchorSide;
+  floatingPosition?: string;
+  fallback: string;
+}) => {
+  let percentage: number | undefined;
+  // This is required when the anchor edge is a logical keyword
+  // (`start/end/self-start/self-end/center`) or a percentage,
+  // not a physical keyword (`top/bottom/left/right`)
+  let axis: 'x' | 'y' | null = getAxis(floatingPosition);
```
jgerigmeyer, line 132: Right now this is only requiring knowledge of the property being declared in cases where the anchor-edge is not a physical keyword -- that should allow us to more quickly support math and custom-var use-cases. However, technically we should be validating that the eventual [property axis matches the anchor axis](https://tabatkins.github.io/specs/css-anchor-position/#anchor-pos):

> The physical <anchor-side> keywords (left, right, top, and bottom) must only be used in inset properties corresponding to the same axis: for example, top and bottom are valid in top and bottom, but not in left and right. Using wrong-axis keywords makes the function invalid.

Similarly, we won't be able to support logical keywords (start/end/center) or percentages in math or custom-var use-cases until we have a way to always determine what the declared property is.

jgerigmeyer:
[transformCSS() - convert positioning to JS](https://trello.com/c/2BVeHDpR/37-transformcss-convert-positioning-to-js)

src/transform.ts
mirisuzanne: This is dependent on writing-mode. In a horizontal writing mode, this will always work. In a vertical writing mode, it would be flipped. (and then `direction` changes the start vs end value)

mirisuzanne (approved):
this looks good, with caveats about writing mode (which I think is a separate story?)

PR #25, opened 2022-09-12
Validating anchor element

sanajaved7:
## Description
- Given a name for an anchor and a name for a floating element (presumably from our parsing step) - this validates whether the anchor element is valid according to the criteria outlined in the spec
- If multiple elements are found with that anchor name, then the first one that is valid is returned along with the floating element from the `validatedForPositioning` function. This is written that this function is what is called before positioning and then positioning uses these elements moving forward. Will return null for the `anchor` and `floating` properties if no valid set found.
- Most of the logic for validation is in the `validAnchorElement` function which goes through the criteria of the spec

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -1,5 +1,7 @@
 import * as csstree from 'css-tree';
 
+import { validatedForPositioning } from './validate.js';
+
 interface DeclarationWithValue extends csstree.Declaration {
   value: csstree.Value;
 }
@@ -28,7 +30,7 @@ export type AnchorSideKeyword =
 export type AnchorSide = AnchorSideKeyword | number;
 
 interface AnchorFunction {
-  anchorEl?: string[];
+  anchorEl?: HTMLElement | null;
   anchorName?: string;
   anchorEdge?: AnchorSide;
   fallbackValue: string;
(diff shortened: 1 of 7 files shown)


src/validate.ts
```diff
@@ -0,0 +1,70 @@
...
+      lastInChain?.style.position === 'absolute' ||
+      getComputedStyle(lastInChain).position === 'absolute'
+    ) {
+      return false;
+    }
+  }
+
+  // el is a descendant of the querying element’s containing block, or the quering element’s containing block is the initial containing block
+  const descendant = floatingElement?.offsetParent?.contains(anchorElement);
+  const floatingCBIsInitialCB = floatingElement?.offsetParent === null;
```
sanajaved7, line 61: @jgerigmeyer @mirisuzanne figuring out the ICB is the part I had the most questions about so I wanted to point out this logic where we check if the `offsetParent` is null. [offsetParent](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) gives us the containing block for an element and I am using it throughout the anchor element validation logic here.

The part that is a bit unexpected/weird is that when an element's containing block is the initial containing block (ICB), `offsetParent` returns null. 

I put together this [CodePen](https://codepen.io/sanajaved/pen/PoezXyb) to help illustrate this a bit more. One part of the HTML is from the initial containing block example from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#example_4) where the example element has a fixed position and thus the CB is the ICB. 

And in the console output, we see what the `offsetParent` value is at various points in the chain and for different examples. At one point I go up through the chain of containing blocks to see if the ultimate containing block does in fact return null.

My reading is that it seems reliable to assume that when `offsetParent` is null, the ICB is the element's containing block.

If this approach makes sense, I think I have two changes still to make. The `offsetParent` [docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) go through the various scenarios when `offsetParent` could be null and it mentions that a `display: none` on the element or its parent could return null. And also that Firefox returns the body element instead of null so I would update the logic to reflect both of those caveats. 

Let me know what you think and if you have other suggestions / thoughts with this approach.

src/validate.ts
```diff
@@ -0,0 +1,70 @@
...
+    anchorElement?.style.position === 'absolute' ||
+    getComputedStyle(anchorElement).position === 'aboslute';
+
+  if (
+    anchorAbsolutelyPositioned &&
+    anchorElement.offsetParent === floatingElement.offsetParent
+  )
+    return false;
+
+  // el has a different containing block from the querying element, the last containing block in el’s containing block chain before reaching the querying element’s containing block is not absolutely positioned
```
oluoluoxenfree, line 39: does this mean you can't anchor something with any absolutely positioned parents, or have i misread it?

oluoluoxenfree (approved):
LGTM; I left one clarifying comment but if i understand correctly, all good.

src/validate.ts
mirisuzanne: That makes sense to me!

src/validate.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  return null;
+}
+
+export function isAbsolutelyPositioned(el: HTMLElement) {
+  return (
+    el.style.position === 'absolute' ||
+    getComputedStyle(el).position === 'absolute'
```
jgerigmeyer, line 28: @sanajaved7 I wonder if we should also include `'fixed'` here?

https://drafts.csswg.org/css-position-3/#absolute-position

jgerigmeyer (approved):
✨

src/parse.ts
```diff
@@ -296,39 +298,46 @@ export function parseCSS(css: string) {
...
       positionFallbacks.forEach((tryBlock) => {
         for (const [prop, value] of Object.entries(tryBlock)) {
           if (typeof value === 'object') {
             const anchorName = (value as AnchorFunction).anchorName;
-            const anchorEl = anchorName ? anchorNames[anchorName] : undefined;
+            const anchorSelectors = anchorName ? anchorNames[anchorName] : [];
+            const anchorEl = validatedForPositioning(
+              floatingEl,
+              anchorSelectors,
+            );
```
jgerigmeyer, line 315: @sanajaved7 This seems to work great!

src/validate.ts
```diff
@@ -0,0 +1,77 @@
...
+  floatingEl: HTMLElement | null,
+  anchorSelectors: string[],
+) {
+  if (!floatingEl) {
+    return null;
+  }
+
+  const anchorElements: NodeListOf<HTMLElement> = document.querySelectorAll(
+    anchorSelectors.join(', '),
+  );
```
jgerigmeyer, line 14: @sanajaved7 I updated this to accept a list of selectors and then just `join` them together. In theory, this means 1) we get them all in tree order, since that's one of the spec requirements, and 2) we can do this once instead of multiple times. Does this seem okay to you?

jgerigmeyer (approved):

src/validate.ts
jgerigmeyer: https://trello.com/c/UkxSJ7Wx

src/validate.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export function hasDisplayNone(el: HTMLElement) {
+  return el.style.display === 'none' || getComputedStyle(el).display === 'none';
+}
+
+// Determines whether the containing block (CB) of the element is the initial containing block (ICB)
+// offsetParent returns null when the CB is the ICB, except in FF where offsetParent returns the body element
+// Excludes elements when they or their parents have display: none
+export function isContainingBlockICB(floatingElement: HTMLElement) {
```
sanajaved7, line 39: @jgerigmeyer @oluoluoxenfree I updated logic for finding the initial containing block here and also added a new test in case y'all would like to take a look and review again before I merge.

jgerigmeyer (approved):

jgerigmeyer:
[Validate anchor element](https://trello.com/c/H2tIXcRc/50-validate-anchor-element)

PR #26, opened 2022-09-13
load polyfill after DOM has loaded

oluoluoxenfree:
## Steps to test/reproduce

## Show me

### DIFF
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -186,7 +186,7 @@ export async function polyfill() {
     // @@@ Wait for DOM load?
     // Or make `autoUpdate` work.
 
-    if (document.readyState) {
+    if (document.readyState === 'complete') {
       position(rules);
     }
     // @@@ update source code
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -183,12 +183,14 @@ export async function polyfill() {
 
   if (Object.values(rules).length) {
     // apply polyfill
-    // @@@ Wait for DOM load?
-    // Or make `autoUpdate` work.
-
-    if (document.readyState === 'complete') {
+    if (document.readyState === 'loading') {
+      document.addEventListener('DOMContentLoaded', () => {
+        position(rules);
+      });
+    } else {
       position(rules);
     }
+
     // @@@ update source code
     // transformCSS(styleData);
   }
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -185,8 +185,10 @@ export async function polyfill() {
     // apply polyfill
     // @@@ Wait for DOM load?
     // Or make `autoUpdate` work.
-    position(rules);
 
+    if (document.readyState) {
+      position(rules);
+    }
     // @@@ update source code
     // transformCSS(styleData);
   }
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -183,8 +183,11 @@ export async function polyfill() {
 
   if (Object.values(rules).length) {
     // apply polyfill
-    if (document.readyState === 'loading') {
-      document.addEventListener('DOMContentLoaded', () => {
+    if (
+      document.readyState === 'loading' ||
+      document.readyState === 'interactive'
+    ) {
+      document.addEventListener('load', () => {
         position(rules);
       });
     } else {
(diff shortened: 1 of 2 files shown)


src/polyfill.ts
```diff
@@ -185,8 +185,10 @@ export async function polyfill() {
     // apply polyfill
     // @@@ Wait for DOM load?
     // Or make `autoUpdate` work.
-    position(rules);
 
+    if (document.readyState) {
```
oluoluoxenfree, line 189: I feel like this should be an event listener but couldn't get one of those versions to work; i guess if autoupdate is going to be used that can cover updates?

jgerigmeyer:
[I want the anchor positioning polyfill to run after the DOM elements are loaded.](https://trello.com/c/zfXmlVeN/53-i-want-the-anchor-positioning-polyfill-to-run-after-the-dom-elements-are-loaded)

src/polyfill.ts
jgerigmeyer: @oluoluoxenfree I agree, it would be better for this to be an event listener -- otherwise, if this check returns falsy, won't the polyfill never run at all?

What event listener(s) did you try, and how were you testing them? The current demo page only applies the polyfill when the button is clicked, which isn't really a good test for this case.

src/polyfill.ts
```diff
@@ -183,9 +183,13 @@ export async function polyfill() {
...
-    // @@@ Wait for DOM load?
-    // Or make `autoUpdate` work.
-    position(rules);
+    if (document.readyState === 'loading') {
+      document.addEventListener('DOMContentLoaded', () => {
+        position(rules);
+      });
+    } else {
+      position(rules);
+    }
```
jgerigmeyer, line 192: @oluoluoxenfree This seems to work? I copied from https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete -- what do you think?

I'm not sure if this is something we can write tests for?

oluoluoxenfree: @jgerigmeyer shouldn't the `else` only fire if the stylesheets are loaded (aka the state is `complete`)? 

I also am not sure about tests; maybe playwright?

jgerigmeyer: > @jgerigmeyer shouldn't the `else` only fire if the stylesheets are loaded (aka the state is `complete`)?

@oluoluoxenfree Hm, I'm not sure... I think we could use the `load` event if we want to wait longer -- but I also assume that `autoUpdate` will handle a shift if needed?

You're probably right though -- better to use `window.addEventListener('load', (event) => { ... })`

src/polyfill.ts
```diff
@@ -183,9 +183,16 @@ export async function polyfill() {
...
   if (Object.values(rules).length) {
     // apply polyfill
-    // @@@ Wait for DOM load?
-    // Or make `autoUpdate` work.
-    position(rules);
+    if (
+      document.readyState === 'loading' ||
+      document.readyState === 'interactive'
+    ) {
+      document.addEventListener('load', () => {
```
jgerigmeyer, line 190: Does this work? I thought it only fires on the `window`:

```suggestion
      window.addEventListener('load', () => {
```

src/polyfill.ts
```diff
@@ -183,9 +183,16 @@ export async function polyfill() {
 
   if (Object.values(rules).length) {
     // apply polyfill
-    // @@@ Wait for DOM load?
-    // Or make `autoUpdate` work.
-    position(rules);
+    if (
+      document.readyState === 'loading' ||
+      document.readyState === 'interactive'
+    ) {
```
jgerigmeyer, line 189:
```suggestion
    if (document.readyState !== 'complete') {
```

src/polyfill.ts
jgerigmeyer: You can test this by adding a new `<script>` tag near the beginning of the `<head>` in `index.html`:

```
    <script>
      console.log(document.readyState);
      document.addEventListener('load', () => {
        console.log('loaded');
      });
    </script>
```

oluoluoxenfree: oh whoops, good catch

jgerigmeyer (approved):

PR #28, opened 2022-09-19
Implement auto-update logic.

jgerigmeyer:


jgerigmeyer:
[I want the anchor positioning to auto-update when elements move after initial load.](https://trello.com/c/5SXcw0Ch/54-i-want-the-anchor-positioning-to-auto-update-when-elements-move-after-initial-load)

src/polyfill.ts
```diff
@@ -118,59 +118,51 @@ export const getPixelValue = ({
...
+                ],
+              });
+            }).then((result) => {
+              const floating: HTMLElement | null =
+                document.querySelector(floatingSel);
+              if (floating) {
+                Object.assign(floating.style, result);
+              } else {
+                cleanup();
+              }
```
jgerigmeyer, line 160: I'm not sure if there's a better way to detect when an element has been removed from the DOM?

src/polyfill.ts
```diff
@@ -118,59 +118,51 @@ export const getPixelValue = ({
...
+                    });
+                    return 0;
+                  }),
+                ],
+              });
+            }).then((result) => {
+              const floating: HTMLElement | null =
+                document.querySelector(floatingSel);
+              if (floating) {
+                Object.assign(floating.style, result);
```
jgerigmeyer, line 157: This is probably inefficient, since we're now updating the inline styles for every property instead of once for every floating element (with all properties), but it was much simpler to implement the watcher like this.

We could also probably be more efficient if we compare the existing value with the new value, and only update inline styles if the value changed -- but that's an optimization I didn't want to get bogged down in right now.

jgerigmeyer (commented):
@sanajaved7 @oluoluoxenfree ~I should probably add some tests here, but~ I think this is functionally complete and ready for review.

I notice that `anchor-scroll` seems to have been removed from the spec entirely, but this PR also means that the scrolling demo works now too.

src/polyfill.ts
```diff
@@ -118,59 +123,51 @@ export const getPixelValue = ({
...
+              } else {
+                cleanup();
```
jgerigmeyer, line 164: Maybe we shouldn't worry about this cleanup? Or we could expose some API for users to call this if/when needed?

sanajaved7: Reading the docs, it seems like `cleanup()` should be called when the floating element is no longer visible, do we need to have a check here for that scenario and then invoke `cleanup()`?

sanajaved7 (approved):
One question about the cleanup step but looks good to me 🎉

tests/e2e/polyfill.test.ts
```diff
@@ -0,0 +1,40 @@
...
+const btnSelector = '#apply-polyfill';
+const floatingSelector = '#my-floating-positioning';
+const anchorSelector = '#my-anchor-positioning';
+
+async function applyPolyfill(page: Page) {
+  const btn = page.locator(btnSelector);
+  return await btn.click();
+}
+
+test('applies polyfill', async ({ page }) => {
```
sanajaved7, line 16: 🎉

src/polyfill.ts
sanajaved7: I think I combined in my question the check for visibility and being removed from the DOM since the docs mentioned both together 😆 but I saw your other comment about checking if an element has been removed from the DOM and from some quick searching, it seems like [Mutation Observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#example) could be an option?

jgerigmeyer: Makes sense -- I opted to remove this for now, and added a new story to look into it.

https://trello.com/c/DTsg8aU0

PR #29, opened 2022-09-20
Add styles

dvdherron:
## Description
- Adds basic styles to anchor-positioning demo. 
- Similar to the [toggles demo ](https://toggles.oddbird.net/#machine)
- uses [Open Props](https://open-props.style/) for some design tokens 

## Steps to test/reproduce

## Show me

image


dvdherron:
[I want a nicer looking and easier to understand demo site.](https://trello.com/c/oTGpE4dw/51-i-want-a-nicer-looking-and-easier-to-understand-demo-site)

dvdherron:
@jgerigmeyer With the polyfill applied, what should the `position-fallback ` version look like?

It currently looks like this:
bild

Also, for the anchor positioning with scroll version, is there a way we can set the scroll position so that the anchor and floats are visible? 

Jonny, I think you mentioned us possibly changing how these are shown (with a dropdown?). Were you thinking of doing that work here or in another story?

jgerigmeyer:
> @jgerigmeyer With the polyfill applied, what should the `position-fallback ` version look like?

I think that changed slightly with the PR I just merged in... Now it looks like this, which is currently "correct" for what's defined:

Screen Shot 2022-09-22 at 3 07 54 PM

> Also, for the anchor positioning with scroll version, is there a way we can set the scroll position so that the anchor and floats are visible?

Yes! Please feel free to edit any of the examples to make them clearer -- I haven't put much thought into them at all.

> Jonny, I think you mentioned us possibly changing how these are shown (with a dropdown?). Were you thinking of doing that work here or in another story?

Your call -- that was just an idea. Maybe a slight preference for that being a new PR?

@dvdherron

jgerigmeyer:
> @jgerigmeyer Is there an attribute that I can hook into when the polyfill is turned on? Say, if we wanted to change the button text to ´Polyfill Applied´?

Done in https://github.com/oddbird/css-anchor-positioning/commit/9b73d0365523011313eb4d04c2db99a680f19ea7

> Yesterday Stacy mentioned that it could possibly be confusing to use "Float" for the element being positioned relative to the anchor. Some might expect these elements to be positioned with the Css property 'float'. I don't know if we need to decide on whether to change this now, but could something like 'Positioned Element' work better?

I'm fine with "Target" or "Positioned" or whatever 👍

jgerigmeyer (approved):

public/demo.css
```diff
@@ -1 +1,202 @@
...
+  font: inherit;
+  font-weight: bold;
+  grid-column: main;
+  max-width: max-content;
+  padding: 0.5em 1em;
+}
+
+[data-button]:hover,
+[data-button]:focus,
+[data-button]:active {
```
stacyk, line 143: @dvdherron when the button becomes disabled after clicking, what would you say about having something like this: 

```
--button-bg-color: transparent;
--button-text: var(--text);
border-color: transparent;
```

Since it probably doesn't need to look like a button since it isn't clickable anymore?

public/demo.css
```diff
@@ -1 +1,202 @@
...
+  padding: 3em 1em;
+}
+
+footer {
+  background: linear-gradient(to bottom right, var(--bg), var(--grad));
+  border-block-start: thin dotted var(--text);
+  display: grid;
+  grid-template-columns: minmax(0, 75ch);
+  padding: 2em 0;
+  justify-content: center;
```
stacyk, line 83: maybe reorder?

public/demo.css
```diff
@@ -1 +1,202 @@
...
+footer p {
+  padding-inline: 1em;
+}
+
+.demo-item {
+  display: grid;
+  grid-template:
+    'heading heading heading' min-content
+    '. elements .' min-content/minmax(0, 5rem) 1fr minmax(0, 5rem);
+  gap: 2rem;
```
stacyk, line 95: Another reorder

stacyk (approved):
I left a few style suggestions but they aren't critical. I didn't review fully for accessibility at this time.

public/demo.css
dvdherron: reordered

public/demo.css
dvdherron: fixed

public/demo.css
dvdherron: Yes. Added disabled styles to this button

image

image

dvdherron:
@stacyk Applied disabled styles to the "Apply Polyfill" button. I think this covers everything if you to give this a final 👍🏽

public/demo.css
```diff
@@ -1 +1,216 @@
...
+}
+
+nav a:any-link,
+nav span {
+  padding: 0.25em;
+}
+
+a:any-link {
+  color: var(--action);
+}
```
stacyk, line 123: @dvdherron I think just one more thing and then we are good here... for this to be accessible while in light mode on the header and footer, the background lime would have to be really light, like #dce9cf. I don't know if it needs the gradient but you can make that call. Everything else looks good.

dvdherron: fixed in https://github.com/oddbird/css-anchor-positioning/pull/29/commits/e330cdc540c3f0eefdc20e62d012c408a1ec0a78

I darkened the text a bit and switched out the gradient in light mode for just one lighter color: 

image

And used a darker gradient in dark mode:
image

PR #30, opened 2022-09-21
Fix bugs with bottom/right positioning.

jgerigmeyer:
- Fixes bugs with `bottom` or `right` positioning (needed to account for width/height of `offsetParent`)
- Uses utility fns from floating-ui instead of `computePosition` api

### DIFF
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -1,14 +1,39 @@
 import {
+  type ElementRects,
+  type FloatingElement,
+  type Platform,
   type Rect,
+  type ReferenceElement,
+  type Strategy,
   autoUpdate,
-  computePosition,
-  offset,
+  platform,
 } from '@floating-ui/dom';
 
 import { fetchCSS } from './fetch.js';
 import { type AnchorPositions, type AnchorSide, parseCSS } from './parse.js';
 
-export const resolveLogicalKeyword = (edge: AnchorSide, isRTL: boolean) => {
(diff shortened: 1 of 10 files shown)


jgerigmeyer:
[Fix bug with bottom/right positioning.](https://trello.com/c/mpbWQy5L/64-fix-bug-with-bottom-right-positioning)

mirisuzanne (commented):
results lgtm

src/polyfill.ts
```diff
@@ -56,65 +81,73 @@ export const getAxisProperty = (axis: 'x' | 'y' | null) => {
...
   }
 
+  const hasPercentage =
+    typeof percentage === 'number' && !Number.isNaN(percentage);
+
+  if (floatingPosition === 'bottom' || floatingPosition === 'right') {
+    offsetParent = getOffsetParent(floatingEl);
+    if (!isElement(offsetParent)) {
+      offsetParent = getDocumentElement(floatingEl);
+    }
```
jgerigmeyer, line 135: @sanajaved7 I figured this was a decent way to deal with the "initial containing block" situation? I don't really know what all this means, but it seems to work, in any case. 🤷 😅

sanajaved7: @jgerigmeyer ya I think this works! Looking at floating's `getDocumentElement` method more closely, the `window.document.documentElement` returns the `html` root element which is the initial containing block.

PR #31, opened 2022-09-22
Where is anchor applied 

oluoluoxenfree:
## Steps to test/reproduce

## Show me


oluoluoxenfree:
[I want to always know the property where an `anchor()` fn is being applied.](https://trello.com/c/Ngg4pSm9/56-i-want-to-always-know-the-property-where-an-anchor-fn-is-being-applied)

src/parse.ts
```diff
@@ -290,6 +303,28 @@ export function parseCSS(css: string) {
...
+        const anchorFnData = customProperties[name];
+        if (anchorFnData) {
+          anchorFunctions[rule.value] = {
+            ...anchorFunctions[rule.value],
+            [this.declaration.property]: anchorFnData,
+          };
+        }
+      }
+    });
+  }
```
jgerigmeyer, line 326: @oluoluoxenfree The goal of the second pass is to find anywhere a custom property is used (e.g. `--center`), and store the inset property it's being used by (e.g. `top`).

I started playing with this just to check if my basic understanding was correct, and I was surprised but I think it actually already works okay as-is 😲 -- but we should test it, and I'm sure there's more cleanup we could do here. Take a look and see what you think.

src/parse.ts
```diff
@@ -171,13 +177,20 @@ function getAnchorNameData(node: csstree.CssNode, rule?: csstree.Raw) {
...
   rule?: csstree.Raw,
 ) {
   if (isAnchorFunction(node) && rule?.value && declaration) {
-    return { [declaration.property]: parseAnchorFn(node) };
+    const data = parseAnchorFn(node);
+    if (declaration.property.startsWith('--')) {
+      customProperties[declaration.property] = data;
+      return;
+    }
+    return { [declaration.property]: data };
```
jgerigmeyer, line 193: We should probably check here to see if `declaration.property` is a valid inset property (top/right/bottom/left).

src/parse.ts
```diff
@@ -290,6 +303,28 @@ export function parseCSS(css: string) {
...
 
+  // Find where CSS custom properties are used
+  if (Object.values(customProperties).length > 0) {
+    csstree.walk(ast, function (node) {
+      const rule = this.rule?.prelude as csstree.Raw | undefined;
+      if (
+        rule?.value &&
+        isVarFunction(node) &&
+        node.children.first &&
+        this.declaration
```
jgerigmeyer, line 314: We should probably also check here to see if `this.declaration.property` is a valid inset property (top/right/bottom/left).

src/parse.ts
oluoluoxenfree: @jgerigmeyer what should we do if it isn't? just return?

src/parse.ts
```diff
@@ -171,13 +177,24 @@ function getAnchorNameData(node: csstree.CssNode, rule?: csstree.Raw) {
...
+    const data = parseAnchorFn(node);
+    if (declaration.property.startsWith('--')) {
+      customProperties[declaration.property] = data;
+      return;
+    }
+    if (isInset(declaration.property)) {
+      return { [declaration.property]: data };
+    } else {
+      return;
+    }
```
jgerigmeyer, line 197:
```suggestion
    }
```

src/parse.ts
```diff
@@ -193,6 +210,16 @@ function getPositionFallbackDeclaration(
...
 }
 
+function isInset(property: string) {
+  const insetProperties: string[] = ['left', 'right', 'top', 'bottom'];
+
+  if (insetProperties.includes(property)) {
+    return true;
+  } else {
+    return false;
+  }
```
jgerigmeyer, line 220:
```suggestion
  const insetProperties = ['left', 'right', 'top', 'bottom'];

  return insetProperties.includes(property);
```

jgerigmeyer (approved):

PR #34, opened 2022-09-27
Parse inline styles

sanajaved7:
## Description
- Updated our fetch logic to call a new function: `fetchInlineStyles`
- This looks for any element with a style attribute that includes the word 'anchor'
- For the selector, it uses either the element's ID but if there is not an ID, then the class list 
- Then it formats the styles with the selector to resemble CSS from style tags so that we can pass it to the parsing function and use it the same way

## Steps to test/reproduce
- In the demo, there is now an example for anchor positioning with inline styles. When the polyfill is applied, it should also correctly anchor the floating and anchor elements with inline styles.


sanajaved7:
[Parse inline styles via style attributes](https://trello.com/c/6erw8wBO/49-parse-inline-styles-via-style-attributes)

src/fetch.ts
```diff
@@ -32,6 +32,28 @@ async function fetchLinkedStylesheets(
...
+// If no elements are found with inline styles, an empty list is returned.
+function fetchInlineStyles() {
+  const elementsWithInlineAnchorStyles =
+    document.querySelectorAll('[style*="anchor"]');
+  const inlineStyles: string[] = [];
+
+  elementsWithInlineAnchorStyles.forEach((el) => {
+    const selector = el.id
+      ? `#${el.id}`
+      : `.${el.classList.value.replaceAll(' ', '.')}`;
```
jgerigmeyer, line 48: Rather than relying on ids or classes, would it be more robust to add a unique selector (class or data-attr or something) directly to the element, then use that? That way we don't need to worry about elements that share the same class.

jgerigmeyer (commented):
@sanajaved7 This works great! Just one minor suggestion for dealing with selectors.

src/fetch.ts
sanajaved7: @jgerigmeyer good thinking - and since ids are unique, maybe we only need to generate and add the selector when the element does not have an id. I guess the selector we add could be an id then too? And then that way we are always using an id as the selector in this block above.

sanajaved7: Yep I was thinking we would only add an id if there wasn't one and that would be a generated unique id. But if you prefer to add a unique selector to all elements (whether or not they have an id) just to rule out any user error that's fine with me too. I'll start with that approach.

jgerigmeyer (approved):
👏

PR #38, opened 2022-10-07
[WIP] Add GH Actions workflow for WPTs

jerivas:
The new workflow runs a subset of WPTs against our polyfill and reports the results to a branch named `<original-branch>--wpt-results`.

Based on the work on https://github.com/GoogleChromeLabs/container-query-polyfill and https://github.com/mehdi-kazemi/scroll-timeline/tree/automate-wpt-test


jerivas:
[Tests (including WPT )](https://trello.com/c/QJVxwJdr/23-tests-including-wpt)

.github/workflows/test-wpt.yml
```diff
@@ -0,0 +1,70 @@
...
+          cache: yarn
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.x'
+      - name: Clone WPT repo
+        uses: actions/checkout@v3
+        with:
+          repository: oddbird/wpt
+          path: wpt
+          ref: master
```
jerivas, line 34: Here we can adjust wpt to come from a different fork or branch

tests/wpt.ts
```diff
@@ -0,0 +1,552 @@
...
+  //   /max\(.*\)/,
+  //   /style\(.*\)/,
+  //   /#container width 399px after padding is applied. #second is removed from the rendering/,
+  //   /ex units/,
+  //   /ch units/,
+  //   /ex relative/,
+  //   /ch relative/,
+];
+
+const CHROME_DEFINITION: BrowserDefinition = {
```
jerivas, line 64: This `*_DEFINITION` constants define the matrix of browsers tested by the workflow. Right now, only a single Chrome version is being tested. Other browsers are commented out but can be enabled

tsconfig.json
```diff
@@ -1,13 +1,15 @@
 {
   "compilerOptions": {
     "module": "node16",
+    "target": "es2022",
```
jerivas, line 4: I had to add this to use top-level await in `wpt.ts`

jerivas:
I'm not sure how to interpret the result number: https://github.com/oddbird/css-anchor-positioning/blob/7308a3723ab936f28fd70c51c45f075afd6ecfd5/test-results/2022-10-06_23-07-00.json#L10-L13

I've reached out to the folks at Google for clarification

jerivas:
Oh, it looks like this is being actively worked on on the CQ repo (committed 4 hours ago) https://github.com/GoogleChromeLabs/container-query-polyfill/commit/0eff67e611571aa96ee3bc7e59ebece98b64114b

tsconfig.json
```diff
@@ -1,13 +1,15 @@
...
     "moduleResolution": "Node16",
     "strict": true,
     "esModuleInterop": true,
     "isolatedModules": true,
     "forceConsistentCasingInFileNames": true,
     "declaration": true,
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "sourceMap": true
```
jgerigmeyer, line 12: I don't think this has any impact, since we're not using TypeScript to generate any JS files (only `.d.ts` files)?

.github/workflows/test-wpt.yml
```diff
@@ -0,0 +1,70 @@
...
+        env:
+          RESULTS_BRANCH: ${{ github.ref_name }}--wpt-results
+        run: |
+          mv test-results test-results.html /tmp
+          git fetch
+          BRANCH_EXISTS=$(git ls-remote --heads origin $RESULTS_BRANCH | wc -l)
+          if [ "$BRANCH_EXISTS" == "1" ]; then git switch $RESULTS_BRANCH; else git switch -c $RESULTS_BRANCH; fi
+          rm -rf test-results test-results.html
+          mv /tmp/test-results /tmp/test-results.html .
+          git add test-results test-results.html
```
jgerigmeyer, line 66: We should figure out how to interpret these results 😅

jerivas:
> Where do the runner.html and wpt.ts files come from?

@jgerigmeyer they come from https://github.com/GoogleChromeLabs/container-query-polyfill and https://github.com/mehdi-kazemi/scroll-timeline/tree/automate-wpt-test. I'm guessing we will need to keep them up to date but only if we want to change how tests are run and reported. At this point we should also be able to contribute back any changes that improve the workflow

tsconfig.json
jerivas: I think that was required to run the TS debugger in VS Code, but yes, we can get rid of it

.github/workflows/test-wpt.yml
jerivas: I got this explanation today

>  the results of this format is just `[numPassingTests, numFailingTests]`.

Is that good enough for us to start?

jgerigmeyer: Seems like a good start -- though we'll quickly want a way to see _which_ tests are failing, so we can focus in on the anchor-positioning ones.

jerivas: Oh yes, I forgot to add that we are running only a subset of tests right now. In our case it's only tests inside `css/css-anchor-position`. See https://github.com/oddbird/css-anchor-positioning/blob/6329519b3e891eec4347e2a8a0aabf1a6ef3fb46/tests/wpt.ts#L49

jgerigmeyer: Do we get any info about _which_ tests are failing though?

jerivas: I'm looking into that now. The JS in `runner.html` is generating the results, so I'm trying to see if we can extract more data from the iframe messages being passed to it

jgerigmeyer: Regardless, this is a good first step. 👍  Ideally (as a developer) I'd love a quick-ish way to see which tests are passing, and/or to run them locally.

jerivas: Yeah, it definitely feels incomplete without that

jerivas:
@jgerigmeyer looks like https://github.com/oddbird/css-anchor-positioning/commit/6329519b3e891eec4347e2a8a0aabf1a6ef3fb46 broke `yarn test:wpt`. If modfying the top-level tsconfig is not an option, perhaps the command in `package.json` can be modified to read the inner tsconfig? Not sure how to do that though 😬

jgerigmeyer:
> If modfying the top-level tsconfig is not an option, perhaps the command in `package.json` can be modified to read the inner tsconfig? Not sure how to do that though 😬

@jerivas Sorry! Fixed in https://github.com/oddbird/css-anchor-positioning/commit/1182e23a1832354570e53060016bd6dc12e7805e

Since we're only using TS to generate types (not to generate transpiled JS), this shouldn't have any practical impact on the distributed code. 👍

tests/runner.html
```diff
@@ -0,0 +1,70 @@
...
+        });
+      }
+
+      function timeout(ms) {
+        return new Promise((resolve) => {
+          setTimeout(resolve, ms);
+        });
+      }
+
+      async function runCqTests(tests) {
```
sanajaved7, line 32: I'm guessing `Cq` here is referring to container queries? If so, we probably want to update that

jerivas:
Okay, so there's a lot of data to go over here. When [`runner.html` completes](https://github.com/oddbird/css-anchor-positioning/blob/09fe929cf4bcb4a0a312fe1b80af83494da08f48/tests/runner.html#L63) if produces a [big JSON array](https://gist.github.com/jerivas/5920711e97cdf65b297ec5d37de41962) with detailed data of each html file, down to the individual test results in each page. Then `wpt.ts` [goes over the results](https://github.com/oddbird/css-anchor-positioning/blob/09fe929cf4bcb4a0a312fe1b80af83494da08f48/tests/wpt.ts#L483-L500) and reduces them down to fail/pass numbers.

I think what we actually want is a way to visualize that big JSON file, just how [wpt.fyi does it](https://wpt.fyi/results/css/css-anchor-position?label=master&label=experimental&aligned). I'm thinking of creating an html file that loads the JSON and uses a small templating library to produce an output similar to wpt.fyi.

This will need to wait until after I'm back from DjangoCon (Oct 24), so I wanted to present the idea in case someone else wants to pick it up in the meantime.

jerivas:
This link should update itself with the latest results for this branch: https://gh-actions-wpt-wpt-results--anchor-position-wpt.netlify.app/

Once we merge with main the results will be available at https://anchor-position-wpt.netlify.app/

tests/report.ts
```diff
@@ -0,0 +1,145 @@
...
+          summary: total === undefined ? [-1, -1] : [passed, total],
+        };
+        byPath[path] ? byPath[path].push(data) : (byPath[path] = [data]);
+      });
+    });
+  });
+
+  // Render the HTML report
+  // Each test gets a row, and the columns are the browser versions
+  const tableHtml = `
```
mirisuzanne, line 59: Doesn't need to hold anything up here, but might be nice to split this out into a stand-alone html template at some point. Could also do some basic styling at that point.

tests/report.ts
```diff
@@ -0,0 +1,145 @@
...
+  <!DOCTYPE html>
+  <html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>WPT Report</title>
```
mirisuzanne, line 64: Include the feature name in the title?

tests/report.ts
```diff
@@ -0,0 +1,145 @@
...
+    <style>
+      td {border: 1px solid}
+      .test-name {display: flex; justify-content: space-between; gap: 1em}
+      .test-bar {color: white; background-color: maroon}
+    </style>
+  </head>
+  <body>
+  <p>Generated at: ${timeStamp}</p>
```
mirisuzanne, line 72: Add an `h1` with the feature name? Also a general name/link to the polyfill repo?

tests/report.ts
```diff
@@ -0,0 +1,145 @@
...
+    wptRepo && wptCommit
+      ? `<p>
+          WPT commit:
+          <a target="_blank" href="https://github.com/${wptRepo}/commit/${wptCommit}">
+            ${wptCommit}
+          </a>
+        </p>`
+      : ''
+  }
+  <br><a href="history.html">History</a>
```
mirisuzanne, line 93: Use a list instead of paragraphs and a free-floating link with a line-break?

tests/report.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+export interface TestPathMap {
+  [key: string]: VersionResult[];
+}
+
+const localDomain = 'http://web-platform.test:8000/';
+
+function gradientBar(active: number, total: number): string {
+  const stop = `${((active / total) * 100).toFixed(0)}%`;
+  return `style="background-image: linear-gradient(to right, darkgreen ${stop}, maroon ${stop})"`;
```
mirisuzanne, line 17: We could make this less repetitive with a global pattern:

```css
td[style*='--result:'] {
  background-image: linear-gradient(to right, darkgreen var(--result), maroon var(--result));
}
```

Then the individual `gradientBar()` output for each result could be simplified to:

```js
return `style="--result:${stop};"`
```

mirisuzanne (approved):
Nice! I left a few comments on ways we could clean up the generated output, but I don't think any of those changes are urgent.

tests/report.ts
jerivas: I agree, this approach is not friendly to collaboration. Do you or @jgerigmeyer have a preferred template language? Would that be nunjucks?

mirisuzanne: Nunjucks has been our default. I've been wondering if we should try out liquid, which is very similar but more actively maintained. (I was strongly encouraged to switch by the creator of 11ty 😅)

jgerigmeyer: As someone responsible for maintaining Nunjucks for a number of years, I would personally not recommend that anyone use it for a new project.

jgerigmeyer (approved):

tests/report.ts
```diff
@@ -0,0 +1,145 @@
...
+                  ${
+                    wptRepo && wptCommit
+                      ? `<a target="_blank" href="https://github.com/${wptRepo}/blob/${wptCommit}/${testPath}">
+                          ${testPath}
+                        </a>`
+                      : testPath
+                  }
+                  <span>
+                    <a target="_blank" href="https://wpt.live/${testPath}" title="Open in wpt.live">🌐</a>
+                    <a target="_blank" href="http://${localDomain}${testPath}" title="Open locally">🏠</a>
```
jgerigmeyer, line 122:
```suggestion
                    <a target="_blank" href="${localDomain}${testPath}" title="Open locally">🏠</a>
```

jgerigmeyer (commented):
@jerivas We still/again have an issue where the reporting of test results as they run (which adds elements to the DOM) is causing tests to fail.

When each test starts running, the [testharness.js](https://github.com/oddbird/wpt/blob/master/resources/testharness.js) file calls `resolve_log()`, which adds a status report (e.g. `<div id="log">Running, 0 complete, 1 remain</div>`) to the DOM. Because the `<body>` _is_ the `#container`, the addition of that new div requires re-calculating the positions, but the test assertion runs too quickly and the test fails.

I think the easiest fix would be adding a wrapper `<div id="container">` around the `<body>` contents in each test (which we [discussed earlier](https://oddbird.slack.com/archives/C03K136KX2A/p1663883114206289?thread_ts=1663872431.086989&cid=C03K136KX2A)), or adding a wrapper around the test results, or an option to wait and only add test results _after_ the tests are run. (Trying locally, any of those cause the tests to pass.)

I'm leaving this PR approved because I'm fine merging it whenever, but I really hope we can get this fixed before we demo the results chart to Mason again 😅

jgerigmeyer:
Not sure how this happened, but seems like it was a fluke...

Screenshot 2022-11-23 at 10 32 53 AM

https://gh-actions-wpt-wpt-results--anchor-position-wpt.netlify.app/2022-11-23t02-17-29.894z

jerivas:
> Not sure how this happened, but seems like it was a fluke...

@jgerigmeyer yeah, I sometimes get random failures where some versions are missing the results array completely, so I wanted to be clear in the report about that. I set it to `0 / -1` in bf90883dd876bf468cbf3946d6ab32b1ed951c44 to make sure the cell is marked as not passing at least

jerivas:
@jgerigmeyer @mirisuzanne Right now we are hard coding the list of browsers we want to test against. Would it help to use a browserslist query to decide that instead? For example, [`last 3 versions and > 0.1% and not dead`](https://browsersl.ist/#q=last+3+versions+and+%3E+0.1%25+and+not+dead) seems like a good starting point. The browser version would be auto-updated each time the report is generated or every time we run dependency upgrades.

jerivas:
> When each test starts running, the [testharness.js](https://github.com/oddbird/wpt/blob/master/resources/testharness.js) file calls resolve_log(), which adds a status report (e.g. `<div id="log">Running, 0 complete, 1 remain</div>`) to the DOM.

Sounds like we need to start adjusting the WPTs themselves. I started doing that on the [`anchor-position-polyfill` branch](https://github.com/web-platform-tests/wpt/compare/master...oddbird:wpt:anchor-position-polyfill) and I'm seeing [a lot more green](https://gh-actions-wpt-wpt-results--anchor-position-wpt.netlify.app/2022-11-24t00-12-52.812z) since fcd65528ae4352e9388c59f1172a87e4dbf6d6fd

jgerigmeyer:
> Would it help to use a browserslist query to decide that instead? For example, [`last 3 versions and > 0.1% and not dead`](https://browsersl.ist/#q=last+3+versions+and+%3E+0.1%25+and+not+dead) seems like a good starting point. The browser version would be auto-updated each time the report is generated or every time we run dependency upgrades.

Makes sense to me. I'm fine with your suggestion, or `defaults` for a simpler approach.

> Sounds like we need to start adjusting the WPTs themselves. I started doing that on the [`anchor-position-polyfill` branch](https://github.com/web-platform-tests/wpt/compare/master...oddbird:wpt:anchor-position-polyfill) and I'm seeing [a lot more green](https://gh-actions-wpt-wpt-results--anchor-position-wpt.netlify.app/2022-11-24t00-12-52.812z) since [fcd6552](https://github.com/oddbird/css-anchor-positioning/commit/fcd65528ae4352e9388c59f1172a87e4dbf6d6fd)

Looks good! Is [this line](https://github.com/web-platform-tests/wpt/compare/master...oddbird:wpt:anchor-position-polyfill#diff-b8002771420c10969dc9ed103bcf6d7efb79165532c02a97b0b19218b21a0bc0R30) doing anything?

jerivas:
> Looks good! Is [this line](https://github.com/web-platform-tests/wpt/compare/master...oddbird:wpt:anchor-position-polyfill#diff-b8002771420c10969dc9ed103bcf6d7efb79165532c02a97b0b19218b21a0bc0R30) (data-anchor-polyfill) doing anything?

@jgerigmeyer (sorry, forgot to reply) yes, it seems like a bunch of tests fail if they find extraneous data attributes 🤷‍♂️

PR #47, opened 2022-11-01
Fix bugs with unknown anchor names

andreubotella:
## Description
It used to be the case that an unknown anchor name would throw an exception. This change fixes that.

## Steps to test/reproduce
Use the `anchor()` function with an anchor name that hasn't been defined anywhere. For example, `left: anchor(--test right)`, with no element having `anchor-name: --test`.

## Show me

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -344,10 +344,9 @@ export function parseCSS(css: string) {
         for (const [prop, value] of Object.entries(tryBlock)) {
           if (typeof value === 'object') {
             const anchorName = (value as AnchorFunction).anchorName;
-            const anchorSelectors =
-              anchorName && anchorName in anchorNames
-                ? anchorNames[anchorName]
-                : [];
+            const anchorSelectors = anchorName
+              ? anchorNames[anchorName] ?? []
+              : [];
             const anchorEl = validatedForPositioning(targetEl, anchorSelectors);
             (tryBlock[prop] as AnchorFunction).anchorEl = anchorEl;
           }
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -344,7 +344,10 @@ export function parseCSS(css: string) {
         for (const [prop, value] of Object.entries(tryBlock)) {
           if (typeof value === 'object') {
             const anchorName = (value as AnchorFunction).anchorName;
-            const anchorSelectors = anchorName ? anchorNames[anchorName] : [];
+            const anchorSelectors =
+              anchorName && anchorName in anchorNames
+                ? anchorNames[anchorName]
+                : [];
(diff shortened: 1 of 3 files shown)


src/parse.ts
```diff
@@ -344,7 +344,10 @@ export function parseCSS(css: string) {
         for (const [prop, value] of Object.entries(tryBlock)) {
           if (typeof value === 'object') {
             const anchorName = (value as AnchorFunction).anchorName;
-            const anchorSelectors = anchorName ? anchorNames[anchorName] : [];
+            const anchorSelectors =
+              anchorName && anchorName in anchorNames
+                ? anchorNames[anchorName]
+                : [];
```
jgerigmeyer, line 350:
```suggestion
            const anchorSelectors = anchorName
              ? anchorNames[anchorName] ?? []
              : [];
```

src/parse.ts
```diff
@@ -361,9 +364,10 @@ export function parseCSS(css: string) {
     const targetEl: HTMLElement | null = document.querySelector(targetSel);
     for (const [targetProperty, anchorObj] of Object.entries(anchorFns)) {
       // Populate `anchorEl` for each `anchor()` fn
-      const anchorSelectors = anchorObj.anchorName
-        ? anchorNames[anchorObj.anchorName]
-        : [];
+      const anchorSelectors =
+        anchorObj.anchorName && anchorObj.anchorName in anchorNames
+          ? anchorNames[anchorObj.anchorName]
+          : [];
```
jgerigmeyer, line 370:
```suggestion
      const anchorSelectors = anchorObj.anchorName
        ? anchorNames[anchorObj.anchorName] ?? []
        : [];
```

jgerigmeyer: Or:

```suggestion
      const anchorSelectors =
        anchorObj.anchorName && Object.hasOwn(anchorNames, anchorObj.anchorName)
          ? anchorNames[anchorObj.anchorName]
          : [];
```

src/validate.ts
```diff
@@ -5,7 +5,7 @@ export function validatedForPositioning(
   targetEl: HTMLElement | null,
   anchorSelectors: string[],
 ) {
-  if (!targetEl) {
+  if (!targetEl || anchorSelectors.length == 0) {
```
jgerigmeyer, line 8:
```suggestion
  if (!targetEl || anchorSelectors.length === 0) {
```

jgerigmeyer (changes requested):
Thanks for your contribution! This was on our radar, but we hadn't gotten to it yet. 👍

A few minor suggested tweaks, then this is good to go. 🚀

jgerigmeyer (approved):

PR #49, opened 2022-11-09
Support specifying anchor name via CSS custom property.

### LINKED ISSUE #45: The `anchor()` function doesn't support variables as anchor targets
The optional first argument of the `anchor()` function, which is a dashed ident that indicates the anchor target, can be replaced by a custom property: for example, `anchor(var(--anchor) left)`. This variable substitution is not supported in this polyfill.

https://codepen.io/andreubotella/pen/mdKexxX

This polyfill doesn't set any additional styles on the target element, and therefore it shows up below the anchor. Whereas on (canary) Chromium with the `CSSAnchorPosition` feature enabled, it is positioned just right of the "Anchor" text.

Supporting this, together with #44, will allow defining a reusable class for tooltips in a CSS file and allow individual elements to define their anchor names or the anchor to be positioned against using `style` attributes, since this is arguably semantic CSS.

jgerigmeyer:
Fixes #45.

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -36,6 +36,7 @@ interface AnchorFunction {
   anchorName?: string;
   anchorEdge?: AnchorSide;
   fallbackValue: string;
+  customPropName?: string;
 }
 
 // `key` is the property being declared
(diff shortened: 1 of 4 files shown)


src/parse.ts
```diff
@@ -239,6 +247,22 @@ function getPositionFallbackRules(node: csstree.CssNode) {
   return {};
 }
 
+function getCSSPropertyValue(el: HTMLElement, prop: string) {
+  return getComputedStyle(el).getPropertyValue(prop);
```
andreubotella, line 251:
```suggestion
  return getComputedStyle(el).getPropertyValue(prop).trim();
```

It seems like different browsers serialize spaces in custom properties in different ways. Here any spaces at the start and at the end of the value shouldn't have any effect, so we trim them.

sanajaved7 (approved):

PR #51, opened 2022-11-16
Set positions via CSS custom properties instead of directly on target elements.

jgerigmeyer:
## Description

Before, we were calculating positions and then setting them directly via `targetElement.style.setProperty(...)`. This isn't ideal for two reasons:

1. It ignores the cascade. What if the `anchor()` call from the stylesheet is overridden by a different set of styles?
2. It doesn't allow any math outside of the `anchor` function, e.g. `top: calc(anchor(--my-anchor 100%) - 50px);`.

With the changes in this PR, we are now replacing uses of `anchor()` with a unique generated CSS custom property directly in the stylesheet, and then after we calculate the position value we simply assign that pixel value to the relevant CSS custom property. This way we can allow CSS itself to handle the cascade, and any relevant math.

## Steps to test/reproduce

View the demo and confirm that the examples still work (including the new "used in math function" example).

Screen Shot 2022-11-16 at 4 30 08 PM

### DIFF
--- a/index.html
+++ b/index.html
@@ -397,7 +397,7 @@
 }</pre
       >
     </section>
-    <!-- <section id="math-function" class="demo-item">
+    <section id="math-function" class="demo-item">
       <h2>
         <a href="#math-function" aria-hidden="true">🔗</a>
         Anchor() [used in math function]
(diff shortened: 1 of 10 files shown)


jgerigmeyer:
[Allow CSS to apply positioning by updating stylesheets instead of applying styles directly.](https://trello.com/c/f1L7Ti8m/57-allow-css-to-apply-positioning-by-updating-stylesheets-instead-of-applying-styles-directly)

jgerigmeyer:
[transformCSS() - anchor() function in math functions](https://trello.com/c/YAl95oDi/26-transformcss-anchor-function-in-math-functions)

jgerigmeyer:
@mirisuzanne @jerivas Since Sana and Olu are out, would the two of you be willing to review this PR sometime? Not urgent, of course.

jerivas (approved):
This looks really hard but works really well 👍 Glad the polyfill now leverages the cascade

mirisuzanne (approved):
Love this approach, and the functionality looks great. Nice work! I remember some discussion about shorter UUIDs in the custom props? That's not urgent, but it would look nicer.

jgerigmeyer:
> I remember some discussion about shorter UUIDs in the custom props? That's not urgent, but it would look nicer.

Yep! That's coming in the next PR... #52

PR #52, opened 2022-11-16
Fix edge cases with CSS custom properties.

jgerigmeyer:


jgerigmeyer:
[Handle cases where there are multiple CSS custom properties with the same name.](https://trello.com/c/UGEMTfVc/66-handle-cases-where-there-are-multiple-css-custom-properties-with-the-same-name)

jgerigmeyer:
[When an anchor fn is applied to a custom property, trace usage even if that prop is then applied to another custom property.](https://trello.com/c/yOP9vqxZ/65-when-an-anchor-fn-is-applied-to-a-custom-property-trace-usage-even-if-that-prop-is-then-applied-to-another-custom-property)

jgerigmeyer:
[transformCSS() - anchor() function in math functions](https://trello.com/c/YAl95oDi/26-transformcss-anchor-function-in-math-functions)

jgerigmeyer:
@mirisuzanne I think I got it to work for an arbitrary number of custom property re-assignments/pass-throughs! I still need to update some tests, but take a look at the new demos and see what you think. 🙏

mirisuzanne:
@jgerigmeyer impressive! Looks good to me.

mirisuzanne (approved):

jgerigmeyer:
@sanajaved7 There's a lot here, and I'm sure this could be cleaner. I'd appreciate your review, and it might be most efficient to walk through the changes together?

sanajaved7:
@jgerigmeyer still catching up so it would definitely help to review together!

jgerigmeyer:
> @jgerigmeyer still catching up so it would definitely help to review together!

@sanajaved7 Makes sense -- propose a time on the calendar?

sanajaved7 (approved):
Thanks for tackling this! This looks good to me and I feel good about the follow-up plan to separate out the custom property parsing logic to share with other polyfills 🎉

sanajaved7:
Test comment ?

PR #61, opened 2022-12-15
Handling fixed position and refactoring validation 

sanajaved7:
## Description
- Refactoring validation to use floating-ui's [getOffsetParent](https://github.com/floating-ui/floating-ui/blob/8670939a7f72587e53d726b481570be0e3b59253/packages/dom/src/utils/getOffsetParent.ts#L39) method to get the correct containing block 
- Notable differences with floating-ui methods: they are asynchronous and the `window` object is returned in certain paths so we updated our logic, specifically when looking for the initial containing block and going up the containing block chain to account for this) 
- For testing this new asynchronous approach, I reworked the tests to be able to use the floating-ui methods and give enough time to wait for the validation to take place


src/parse.ts
```diff
@@ -739,10 +742,13 @@ export function parseCSS(styleData: StyleData[]) {
     if (positionFallbacks) {
       const targetEl: HTMLElement | null = document.querySelector(targetSel);
       // Populate `anchorEl` for each fallback `anchor()` fn
-      positionFallbacks.forEach((tryBlock) => {
+      positionFallbacks.forEach(async (tryBlock) => {
```
jgerigmeyer, line 745: Minor note, but my understanding is that using an `async` cb in `forEach` is generally not a good idea -- I'd recommend switching this to a `for...of` loop instead.

src/validate.ts
```diff
@@ -42,31 +44,27 @@ export function hasDisplayNone(el?: HTMLElement | null) {
 // - `offsetParent` returns `null` when the CB is the ICB,
 //   except in Firefox where `offsetParent` returns the `body` element
 // - Excludes elements when they or their parents have `display: none`
```
jgerigmeyer, line 46: Comments are out of date.

src/validate.ts
```diff
@@ -42,31 +44,27 @@ export function hasDisplayNone(el?: HTMLElement | null) {
...
-    navigator.userAgent.includes('Firefox');
-
-  const offsetParentNullOrBody =
-    targetElement.offsetParent === null || cbIsBodyElementFromFF;
-
-  if (offsetParentNullOrBody && !isDisplayNone) {
-    return true;
-  }
-  return false;
+export async function isContainingBlockICB(
```
jgerigmeyer, line 47:
```suggestion
export function isContainingBlockICB(
```

src/validate.ts
```diff
@@ -42,31 +44,27 @@ export function hasDisplayNone(el?: HTMLElement | null) {
...
+
   // If el has the same containing block as the querying element,
   // el must not be absolutely positioned:
+  // A separate check for fixed positioning is added here because it's offsetParent will always resolve to null: https://w3c.github.io/csswg-drafts/cssom-view/#extensions-to-the-htmlelement-interface
   if (
-    isAbsolutelyPositioned(anchor) &&
-    anchor.offsetParent === target.offsetParent
+    (isAbsolutelyPositioned(anchor) &&
+      anchorContainingBlock === targetContainingBlock) ||
+    isFixedPositioned(anchor)
```
jgerigmeyer, line 67: So if an anchor has `position: fixed`, this means it will never be considered a valid anchor?

src/validate.ts
```diff
@@ -75,27 +73,34 @@ export function isValidAnchorElement(anchor: HTMLElement, target: HTMLElement) {
...
   }
 
-  // Either el must be a descendant of the querying element's containing block,
-  // or the querying element's containing block must be
-  // the initial containing block:
-  const isDescendant = Boolean(target.offsetParent?.contains(anchor));
-  const targetCBIsInitialCB = isContainingBlockICB(target);
+  let isDescendant;
+  if (targetContainingBlock && targetContainingBlock === window) {
+    isDescendant = (targetContainingBlock as Window).document.contains(anchor);
```
jgerigmeyer, line 98: Could we potentially use `platform.getDocumentElement()` here? Or maybe that's not helpful.

src/validate.ts
```diff
@@ -42,31 +44,27 @@ export function hasDisplayNone(el?: HTMLElement | null) {
...
+export async function isValidAnchorElement(
+  anchor: HTMLElement,
+  target: HTMLElement,
+) {
+  const anchorContainingBlock = await platform.getOffsetParent?.(anchor);
+  const targetContainingBlock = await platform.getOffsetParent?.(target);
+
   // If el has the same containing block as the querying element,
   // el must not be absolutely positioned:
+  // A separate check for fixed positioning is added here because it's offsetParent will always resolve to null: https://w3c.github.io/csswg-drafts/cssom-view/#extensions-to-the-htmlelement-interface
```
jgerigmeyer, line 63:
```suggestion
  // el must not be absolutely positioned.
  // A separate check for fixed positioning is added here
  // because its offsetParent will always resolve to null:
  // https://w3c.github.io/csswg-drafts/cssom-view/#extensions-to-the-htmlelement-interface
```

src/validate.ts
sanajaved7: Going to check if we still need this now that we're not using the built-in offsetParent

sanajaved7:
[I want support for `fixed` positioned elements.](https://trello.com/c/UkxSJ7Wx/61-i-want-support-for-fixed-positioned-elements)

src/validate.ts
sanajaved7: @jgerigmeyer I started making this change but I ran into some type weirdness when the the `targetContainingBlock` is a window. Looking at the type locally in my editor(for `platform.getDocumentElement`) I see: `(property) Platform.getDocumentElement?: ((element: Element) => Promisable<HTMLElement>) | undefined` which makes sense why it's rejecting the `Window` types but that seems to not match up to the types [here](https://github.com/floating-ui/floating-ui/blob/ae2e56e751d247d81b162112cd46238c5290e685/packages/dom/src/utils/getDocumentElement.ts#L3). Have you run into something like this before?

src/validate.ts
```diff
@@ -1,5 +1,14 @@
 import { platform } from '@floating-ui/dom';
 
+import { getCSSPropertyValue } from './parse.js';
+
+function hasStyle(element: HTMLElement, cssProperty: string, value: string) {
+  return (
+    element.style.getPropertyValue(cssProperty) === value ||
```
jgerigmeyer, line 7: I wonder if we should `trim()` the result here -- I've read that browsers differ in how they return values with possible trailing whitespace.

jgerigmeyer: Also, remind me why this (`element.style.getPropertyValue(...)`) is needed in addition to `getComputedStyle(el).getPropertyValue(...)`? How do they differ?

src/validate.ts
```diff
@@ -14,59 +25,47 @@ export function validatedForPositioning(
...
+export function isFixedPositioned(el: HTMLElement) {
+  return hasStyle(el, 'position', 'fixed');
 }
 
-export function hasDisplayNone(el?: HTMLElement | null) {
+export function isAbsolutelyPositioned(el?: HTMLElement | null) {
   return Boolean(
     el &&
-      (el.style.display === 'none' || getComputedStyle(el).display === 'none'),
+      (hasStyle(el, 'position', 'fixed') ||
```
jgerigmeyer, line 43:
```suggestion
      (isFixedPositioned(el) ||
```

src/validate.ts
```diff
@@ -75,27 +74,37 @@ export function isValidAnchorElement(anchor: HTMLElement, target: HTMLElement) {
...
-    while (currentCB && currentCB !== target.offsetParent) {
+  if (anchorContainingBlock !== targetContainingBlock) {
+    let currentCB: Element | Window | undefined;
+    const anchorCBchain: typeof currentCB[] = [];
+
+    currentCB = anchorContainingBlock;
+    while (
+      currentCB &&
+      currentCB !== targetContainingBlock &&
+      currentCB != window
```
jgerigmeyer, line 85: Is it intentional to use `!=` instead of `!==` here?

src/validate.ts
jgerigmeyer: Yep, it looks like the `types` file is out of date there. Not a big deal, this is fine as-is. 👍

src/validate.ts
```diff
@@ -14,59 +25,47 @@ export function validatedForPositioning(
...
-
-  const offsetParentNullOrBody =
-    targetElement.offsetParent === null || cbIsBodyElementFromFF;
-
-  if (offsetParentNullOrBody && !isDisplayNone) {
-    return true;
-  }
-  return false;
+// is the initial containing block (ICB)
+export function isContainingBlockICB(
```
jgerigmeyer, line 50: I wonder if it would be clearer to rename this e.g. `isWindow` or something?

jgerigmeyer (commented):
@sanajaved7 Nice work! Do you think it's worth adding a fixed-position demo, or can we point to a WPT test to demonstrate that this works? I was hoping this would fix the failing `anchor-position-003.html` test, but it doesn't seem to. 😕

src/validate.ts
sanajaved7: I think we started first with `element.style` but since that only returns inline styles we added computed styles to get styles from stylesheets. Looking at it again, I think `getComputedStyle` factors in both inline and stylesheets so we can remove the `element.style` check ✅

src/validate.ts
jgerigmeyer: Nevermind -- this matches the naming internally 👍

jgerigmeyer (approved):
Looks great -- and thanks for fixing that WPT! 🎉

PR #64, opened 2022-12-16
Run WPTs against browserslist

jerivas:
## Description
Use browserslist queries to get the latest version numbers for each browser. Removes the need to manually update the test file on every browser release (which happens every few weeks). The only exception is Safari, which has a much slower velocity and each version if tied to the underlying iOS/macOS version.

## Steps to test/reproduce
The [WPT report generated by this PR](https://wpt-browserslist-wpt-results--anchor-position-wpt.netlify.app/) includes the most recent browser versions. In contrast, the [main report](https://anchor-position-wpt.netlify.app/) is using some old, hard-coded versions.

PR #65, opened 2022-12-16
Fix automatic WPT comment

jerivas:
## Description
Due to different GH Action events, the old logic didn't work when opening pull requests, and no comment was created. This should fix it.


jerivas:
I'm merging this because the automated comment was created successfully, and this can be reviewed as part of #64

PR #70, opened 2023-01-11
Add support for `anchor-size()` fn.

jgerigmeyer:


jgerigmeyer:
[anchor-size() function](https://trello.com/c/ESBQoj2Z/25-anchor-size-function)

src/polyfill.ts
```diff
@@ -170,10 +170,10 @@ export const getPixelValue = async ({
         anchorRect[axis] + anchorRect[dir] * ((percentage as number) / 100);
       switch (targetProperty) {
         case 'bottom':
-          value = (offsetParent as HTMLElement).clientHeight - value;
+          value = (offsetParent as HTMLElement).offsetHeight - value;
           break;
         case 'right':
-          value = (offsetParent as HTMLElement).clientWidth - value;
+          value = (offsetParent as HTMLElement).offsetWidth - value;
```
jgerigmeyer, line 176: @mirisuzanne I think these changes from `clientWidth` to `offsetWidth` might actually be incorrect -- and I'm also not sure if I need to worry about scaling here. This change seemed to fix one test (first test in https://github.com/web-platform-tests/wpt/blob/6cd6741bc20cd6b3ad9f95d19123b7850c696d79/css/css-anchor-position/anchor-position-inline-004.html), but it also caused other tests with borders to fail (first 2 in https://github.com/web-platform-tests/wpt/blob/7a0b0c285b91a0b1fa08da58c964b82e5cf40733/css/css-anchor-position/anchor-position-borders-001.html).

Do you know what the correct way to convert from a `left` to a `right` px value is? That's the goal here. E.g. given the distance `x` that would the `left` position to the _right_ edge of a rectangle, how to consistently convert that to a `right` positioning value?

jgerigmeyer: When I play with this, it looks like `right: 0px` does not include borders for the containing block -- so I think this should remain `clientWidth` and `clientHeight`... 🤔

mirisuzanne: Yeah, to get the right offset based on the x offset, you need: the width of the reference box, minus the x offset, minus the width of the target.

For the parent context width, the relevant box is the 'content box' (`clientWidth`?), but for the the target element you would need the full 'border box' including borders (`offsetWidth`?).

jgerigmeyer: For now I added a hack in https://github.com/oddbird/css-anchor-positioning/pull/70/commits/2b48e0655ebc78a6a6c725e5a52cb339a67fafa1, but I'm not confident in it.

src/polyfill.ts
```diff
@@ -64,65 +105,109 @@ export const getPixelValue = async ({
...
+        let vertical = false;
+        // Logical keywords require checking the writing-mode
+        // of the target element (or its containing block)
+        if (targetEl) {
+          // `block` and `inline` should use the writing-mode of the element's
+          // containing block, not the element itself:
+          // https://trello.com/c/KnqCnHx3
+          vertical = getCSSPropertyValue(targetEl, 'writing-mode').startsWith(
+            'vertical-',
+          );
```
mirisuzanne, line 129: I think ideally we also handle `sideways-` - it's [experimental](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode#formal_syntax), but people use it. I'm not familiar with `getCSSPropertyValue()`, but I assume it's getting the computed value of the property? I _think_ that should work.

src/polyfill.ts
```diff
@@ -26,6 +32,24 @@ export const resolveLogicalKeyword = (side: AnchorSide, rtl: boolean) => {
...
+  vertical: boolean,
+) => {
+  let resolved: 'width' | 'height' | undefined;
+  switch (size) {
+    case 'block':
+    case 'self-block':
+      resolved = vertical ? 'width' : 'height';
+      break;
+    case 'inline':
+    case 'self-inline':
```
mirisuzanne, line 46: For the `self-` options, do we need to check the both the writing mode of the element and the parent? I'm not clear how that one works, but my guess would be we want to use whichever anchor-element axis is the target-element's block or inline?

src/polyfill.ts
jgerigmeyer: Yes, `getCSSPropertyValue()` is our shorthand fn for `getComputedStyle(el).getPropertyValue(prop).trim()`.

src/polyfill.ts
jgerigmeyer: I think according to the spec it's reversed -- `inline` and `block` should use the writing mode of the element's containing block. I added a story: https://trello.com/c/KnqCnHx3/55-i-want-better-support-for-writing-modes

jgerigmeyer:
@sanajaved7 Ping on this review

src/polyfill.ts
```diff
@@ -64,65 +105,110 @@ export const getPixelValue = async ({
...
+      }
+    }
+
+    const dir = getAxisProperty(axis);
+    if (hasPercentage && axis && dir) {
+      let value =
+        anchorRect[axis] + anchorRect[dir] * ((percentage as number) / 100);
+      switch (targetProperty) {
+        case 'bottom': {
+          let offsetHeight = (offsetParent as HTMLElement).clientHeight;
```
sanajaved7, line 190: Do you think it would be good to surface this workaround for inline elements in the README (or some other docs)? I'm not an expert on `clientWidth` or `clientHeight` at all so I don't know if this approach has any drawbacks but maybe it could be useful context for users just in case. 

And either way, we might want to pull out any notes or things we want to call out about our implementation throughout the polyfill into a single place for users so this could be done later too as a part of that.

sanajaved7 (approved):
Just had one question about docs and notes but looks good! 🎉

src/polyfill.ts
jgerigmeyer: @sanajaved7 What do you think about the new [limitations](https://github.com/oddbird/css-anchor-positioning/tree/anchor-size#limitations) section of the readme?

sanajaved7: Looks great! 💯

PR #74, opened 2023-01-19
Support multiple targets with the same selector but different valid anchors.

jgerigmeyer:


jgerigmeyer:
[Bug: With a single anchor-name, anchor element may be different for each given target element.](https://trello.com/c/MKApHxX7/83-bug-with-a-single-anchor-name-anchor-element-may-be-different-for-each-given-target-element)

src/validate.ts
```diff
@@ -82,7 +78,11 @@ export async function isValidAnchorElement(
     }
     const lastInChain = anchorCBchain[anchorCBchain.length - 1];
 
-    if (lastInChain && isAbsolutelyPositioned(lastInChain as HTMLElement)) {
```
jgerigmeyer, line 85: @sanajaved7 Some new unit tests using jsdom were failing when `lastInChain` was `window`, so I updated this (and modified the `window` check).

src/polyfill.ts
```diff
@@ -253,25 +249,34 @@ function position(rules: AnchorPositions) {
...
           });
+        } else {
+          // Use fallback value
+          const resolved = await getPixelValue({
+            targetProperty: property,
+            anchorSide: anchorValue.anchorSide,
+            anchorSize: anchorValue.anchorSize,
+            fallback: anchorValue.fallbackValue,
+          });
+          root.style.setProperty(anchorValue.uuid, resolved);
```
jgerigmeyer, line 260: Before, we weren't using the fallback value consistently for invalid queries. This fixes that.

sanajaved7 (approved):
This also looks good, thanks for catching and fixing this! 🐛

PR #75, opened 2023-01-20
Better validation of anchor queries.

jgerigmeyer:


jgerigmeyer:
[Validate anchor-query](https://trello.com/c/z86sRAs0/60-validate-anchor-query)

src/validate.ts
```diff
@@ -107,7 +107,13 @@ export async function validatedForPositioning(
   targetEl: HTMLElement | null,
   anchorSelectors: string[],
 ) {
-  if (!targetEl || anchorSelectors.length === 0) {
+  if (
+    !(
+      targetEl instanceof HTMLElement &&
+      anchorSelectors.length &&
+      isAbsolutelyPositioned(targetEl)
```
jgerigmeyer, line 114: @sanajaved7 My understanding is that one of the requirements is for the target to be absolutely positioned: https://drafts.csswg.org/css-anchor-1/#queries

sanajaved7: Yes that sounds right to and just a quick glance at the WPTs shows that most of them have something like `.target {
  position: absolute;
}`

sanajaved7 (approved):

PR #81, opened 2023-02-09
Support @position-fallback.

jgerigmeyer:


jgerigmeyer:
[Handle position-fallback @try blocks.](https://trello.com/c/Xm8bI3as/58-handle-position-fallback-try-blocks)

jgerigmeyer:
@sanajaved7 @mirisuzanne I think this is now complete for basic `@position-fallback` functionality, and ready for code and/or functional review! ✨ 

Remaining items (include new spec changes) are documented at the bottom of the README, and some have Trello cards as well.

mirisuzanne (approved):
as far as I can tell, the functionality looks good to me. If you want code review on my end, it might be helpful to do that with a sync call. It's interesting when I load it in arc (with experimental web platform features), I'm told that I don't need the polyfill - but it sure seems like we support more features than they do at this point. :)

jgerigmeyer:
> It's interesting when I load it in arc (with experimental web platform features), I'm told that I don't need the polyfill - but it sure seems like we support more features than they do at this point.

Interesting! We're testing for support with `'anchorName' in document.documentElement.style`, but maybe we want to be more specific?

mirisuzanne:
> Interesting! We're testing for support with 'anchorName' in document.documentElement.style, but maybe we want to be more specific?

Yeah, but probably not worth the effort for an experimental implementation. We could add that in if some browser actually ships a partial feature?

src/polyfill.ts
```diff
@@ -243,52 +276,118 @@ export const getPixelValue = async ({
...
+            elements: { floating: target },
+            strategy: 'absolute',
+          } as unknown as MiddlewareState,
+          {
+            boundary: offsetParent,
+            rootBoundary: 'document',
+            padding: getMargins(target),
+          },
+        );
+        // If none of the sides overflow, use this `@try` block and stop loop...
```
sanajaved7, line 369: 🎉

sanajaved7 (approved):
Very nice work on this!! 👏🏼 ✨ 🚀

jgerigmeyer:
[Handle for sizing properties in position-fallback @try blocks](https://trello.com/c/DcY9OWZC/46-handle-for-sizing-properties-in-position-fallback-try-blocks)

PR #87, opened 2023-03-02
Allow updating anchor on every animationFrame.

### LINKED ISSUE #73: Update when anchor moves
From what I can tell, the current implementation is responding to container resize/scroll. But, the browser implementation will also let update anchor positioning when the anchor moves within the container. That movement could be triggered in various ways.

See this demo: https://codepen.io/jh3y/pen/BaPwEMz/bb7b743f7935ca5e415629be01556d3a

Dragging the anchor will keep the boat tethered currently where anchoring is supported.

In this demo, the anchor position is dictated by a grid layout: https://codepen.io/jh3y/pen/xxzayVJ

Changing the grid layout works with the API but isn't detected in the polyfill.

There's also another scenario where the anchor could be updated via a dynamic custom property where the change isn't detected or picked up (That may be worth another issue. This demo shows it in action: https://codepen.io/jh3y/pen/LYrxpjB).

Understandably some of these scenarios might be tricky to monitor. Especially in a performant way. Is there perhaps a trade-off to say "This covers the main cases"? Interested to hear your thoughts.

jgerigmeyer:
Fixes #73.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,36 @@ To use the polyfill, add this script tag to your document `<head>`:
 
 You can view a more complete demo [here](https://anchor-polyfill.netlify.app/).
 
+## Configuration
+
+The polyfill accepts one argument (type: `boolean`, default: `false`), which
+determines whether anchor calculations should [update on every animation
+frame](https://floating-ui.com/docs/autoUpdate#animationframe) (e.g. when the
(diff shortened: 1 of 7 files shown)


jerivas (approved):
Not 100% familiar with what's going on, but from what I can gather it all makes sense

sanajaved7 (approved):
🚀 ✨

PR #90, opened 2023-03-07
Adding support for top-layer

### LINKED ISSUE #89: Absolutely positioned anchor
> The problem seems to be absolute position of anchor or any of its ancesors.

I think that might be correct as per the [spec](https://w3c.github.io/csswg-drafts/css-anchor-position/#acceptable-anchor-element)?

> If el has the same [containing block](https://w3c.github.io/csswg-drafts/css-display-4/#containing-block) as query el, el is not [absolutely positioned](https://w3c.github.io/csswg-drafts/css-position-3/#absolute-position).
>
> If el has a different [containing block](https://w3c.github.io/csswg-drafts/css-display-4/#containing-block) from query el, the last containing block in el’s [containing block chain](https://w3c.github.io/csswg-drafts/css-display-4/#containing-block-chain) before reaching query el’s containing block is not [absolutely positioned](https://w3c.github.io/csswg-drafts/css-position-3/#absolute-position).

But I might be missing something. If you think there's a bug in how this polyfill interprets the spec in regard to determining a valid anchor, please open a new issue. 👍

_Originally posted by @jgerigmeyer in https://github.com/oddbird/css-anchor-positioning/issues/84#issuecomment-1456385890_

sanajaved7:
Fixes #89.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ option can be set by setting the value of
 
 This polyfill doesn't (yet) support the following:
 
-- top layer anchor elements
+- Tracking the order of elements in the [top-layer](https://fullscreen.spec.whatwg.org/#new-stacking-layer) to invalidate top-layer target elements from anchoring to succeding top-layer anchors. See [this WPT](https://github.com/web-platform-tests/wpt/blob/master/css/css-anchor-position/anchor-position-top-layer-006.html) for an example.
 - `anchor-default` property
 - `anchor-scroll` property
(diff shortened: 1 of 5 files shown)


jgerigmeyer:
[Support for top layer anchor elements](https://trello.com/c/negVJZ3k/84-support-for-top-layer-anchor-elements)

jgerigmeyer (approved):

Issue #91, opened 2023-03-10
applying the polyfill to dynamic elements


dreyks:
Hey! thx for the polyfill!

I'm trying to bolt it onto a dropdown and I'm applying the polyfil after the dropdown is open. This works. But when I close the dropdown and open it again - the css-vars on the element aren't being set.

As far as I understood from the code since the CSS was already transformed and there are no `anchor()` functions anymore no elements are targeted

is it somehow possible to achieve what i'm trying to achieve? :)

jgerigmeyer:
@dreyks Thanks for your offer! We're always happy for contributions. 😄 

I am planning to take a look at this issue in the next few days, starting by providing an API for users to re-run the polyfill on demand. It's a bit tricky because the polyfill makes changes directly to the original source CSS, so re-running it isn't as simple as just doing the same thing again. But I have some ideas of how to address this. (Then later hopefully we can also add an observer or something so that the polyfill itself can re-run intelligently.)

In this case it would be helpful if you could share the HTML/JS code you're trying to run -- that way as I work on it I can test directly against your use-case (and add it to the [demo page](https://anchor-polyfill.netlify.app/)).

dreyks:
thx for the response!

i'm doing this on a closed-source ui library, i'll try to craft a simplified example

dreyks:
meanwhile, I've moved the definition of `anchorFunctions` outside of `parseCSS` [here](https://github.com/oddbird/css-anchor-positioning/blob/7ae864c5d0de24369cb8d6e40600a95a05ca6ad7/src/parse.ts#L477) so that it is maintained between polyfill invocations and now everything seems to be working, but i'm not sure it's the most performant solution 😂

dreyks:
here's the example (i realized I forgot how to write DOM-oriented JS without React 😂 )

https://codesandbox.io/s/wonderful-darkness-tmslme?file=/index.html

lucsoft:
Any updates on this? Seeing how Chrome will land CSS Anchor Positioning, in May, would be great to have this issue solved :)

jgerigmeyer:
@lucsoft I agree! 😅 

I'm hoping to spend some time on this in April, but honestly we could also use help from the community. Code contributions are welcome, as well as [financial support](https://opencollective.com/oddbird-open-source) (especially from companies that plan to use this in production).

goodwinchris:
Any movement here? Can't really use the polyfill in my angular application without it

[...]

PR #123, opened 2023-09-22
WIP: Allow unapplying and reapplying polyfill.

jgerigmeyer:
This adds a top-level `restore` function, which attempts to un-apply the polyfill's effects. This allows then re-applying the polyfill, if desired.

This may not be the best solution for #91, but it seemed like a reasonable first step, and might have value on its own as well.

### DIFF
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -444,7 +444,7 @@ export async function polyfill(animationFrame?: boolean) {
   // pre parse CSS styles that we need to cascade
   const cascadeCausedChanges = await cascadeCSS(styleData);
   if (cascadeCausedChanges) {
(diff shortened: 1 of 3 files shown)


jgerigmeyer:
[Support for dynamically-added elements.](https://trello.com/c/Uky665HH/86-support-for-dynamically-added-elements)

jamesnw:
> Hi! Do you have any plans to continue or complete this PR? Let me know if there's anything I can do to help.

We'd love to make this happen, but honestly we could also use help from the community. Code contributions are welcome, as well as [financial support](https://opencollective.com/oddbird-open-source) (especially from companies that plan to use this in production).

PR #177, opened 2024-05-09
Update Limitations to match current spec

jamesnw:
## Description
1. Updates the Limitations section in README against the spec as of March. https://www.w3.org/TR/css-anchor-position-1
2. Adds Status note to [Demo](https://deploy-preview-177--anchor-polyfill.netlify.app/), and deprecation notice on `@position-fallback` example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,15 @@ option can be set by setting the value of
 
 ## Limitations
 
-This polyfill doesn't (yet) support the following:
+This polyfill was implemented against an early version of the spec, and updates
+were paused to allow the syntax to solidify. Now that browsers are working on
+implementation, we would like to bring it up to date.
 
-- `position-try`, `position-try-options`, and `position-try-order` properties
+While this polyfill supports many basic use cases, it doesn't (yet) support the
+following:
+
+- The `@position-try` rule
+- The `position-try-options`, `position-try-order`, or `position-try` properties
 - `anchor-scope` property
 - `position-anchor` property
 - `inset-area` property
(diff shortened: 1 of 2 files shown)


jamesnw:
There were also some additions to the spec where I'm not sure if changes are needed-

1. New section on "[anchor relevance](https://www.w3.org/TR/css-anchor-position-1/#anchor-relevance)", but I'm unclear if this requires any changes.
2. `CSSPositionTryRule` DOM interface has been added- can this be polyfilled as well?
3.  In [creating an anchor](https://www.w3.org/TR/css-anchor-position-1/#name), new specification that `Pseudo elements have same implicit anchor as originating element unless otherwise specified`. I'm uncertain if this requires a change
4. New requirements for [acceptable anchor points](https://www.w3.org/TR/css-anchor-position-1/#acceptable-anchor-element):
   1. el is either an [element](https://www.w3.org/TR/css-display-3/#elements), or a [pseudo-element (https://www.w3.org/TR/selectors-4/#pseudo-element) that acts like one.
   2. el is not in the [skipped contents](https://www.w3.org/TR/css-contain-2/#skips-its-contents) of another element.
5. [Anchor function validity](https://www.w3.org/TR/css-anchor-position-1/#anchor-valid)- adds new requirement that there be a target element.

Also, in the existing limitations, I'm seeing "anchor functions assigned to `inset-*` properties or `inset` shorthand  property"-  I think this one no longer applies, as I don't see a spec for it.

jgerigmeyer (approved):

SondraE (approved):
@jamesnw Let's link people to the GH sponsor page instead of the Open Collective page. We are planning to add a specific goal for Anchor Positioning funding on the GH sponsor page. Seems like it may be a good idea to add that goal to this spec demo whenever we have that on GH.
https://github.com/sponsors/oddbird

PR #178, opened 2024-05-15
Support list of anchor names

jamesnw:
## Description
The spec now supports a list of anchor-name values, for instance `anchor-name: --a, --b;`. 

## Related Issue(s)
n/a

## Steps to test/reproduce
Check the [demo page](https://deploy-preview-178--anchor-polyfill.netlify.app/#anchor-name-list), apply the polyfill, and see that two targets can use different anchor names to target the same anchor element.

Also fixes WPT test css/css-anchor-position/anchor-name-004.html.

## Show me
image

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -338,10 +338,12 @@ function getAnchorNameData(node: csstree.CssNode, rule?: csstree.Raw) {
     node.value.children.first &&
     rule?.value
   ) {
-    const name = (node.value.children.first as csstree.Identifier).name;
-    return { name, selector: rule.value };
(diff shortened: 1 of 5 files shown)


src/parse.ts
```diff
@@ -338,10 +338,12 @@ function getAnchorNameData(node: csstree.CssNode, rule?: csstree.Raw) {
     node.value.children.first &&
```
jgerigmeyer, line 338:
```suggestion
    !node.value.children.isEmpty &&
```

jgerigmeyer (approved):
👏

PR #195, opened 2024-05-20
Update acceptable anchor element algorithm

### LINKED ISSUE #103: Support absolutely positioned anchors (not just top layer)
Hi Oddbird,

There's recently a major change to the anchor finding algorithm:
1. An element can use absolutely positioned anchors that come earlier in tree order
2. When there's a name conflict, we should use the last one instead of the first one

Please see https://github.com/w3c/csswg-drafts/issues/8165 (especially comments after https://github.com/w3c/csswg-drafts/issues/8165#issuecomment-1469912952) for more details about this change.

ayoreis:
Hi, I updated the acceptable anchor element algorithm because I needed it for a project but I think it's good to merge upstream.

All test pass (one needed updating) and I added a new example to the home page.

I left a TODO which I already started to fix but decided it would be better to separate into another PR (pseudo-element anchors). 

Closes #103


ayoreis:
Thanks for looking at this @jamesnw. 

Turns out I only ran the unit tests, I'm already working on fixing the e2e tests.

For the Web Platform Tests I'm getting this error: `Error: invariant: WPT_MANIFEST environment variable must be set`. What do I need to set it to?

ayoreis:
Hi, sorry for taking so long. Quick update: I found the issue and started fixing it (Floating UI's `platform.getOffsetParent` does not work to get an element's containing block in all cases), I'll commit the fix tomorrow.

To check for regressions on a WPT run do you just use expectations metadata and then search for `FAIL [expected PASS]` on the output or is there a better way?

jamesnw:
> To check for regressions on a WPT run do you just use expectations metadata and then search for `FAIL [expected PASS]` on the output or is there a better way?

I ran on main, and then on my branch, and exported the results as JSON from both. Then I did a diff between the two, which was incredibly noisy and pretty annoying. I would love to make that easier.

index.html
```diff
@@ -663,6 +664,48 @@ <h2>
...
   right: anchor(--my-anchor-update right);
   top: anchor(--my-anchor-update bottom);
+}</code></pre>
+    </section>
+    <section id="anchor-absolute" class="demo-item">
+      <h2>
+        <a href="#anchor-absolute" aria-hidden="true">🔗</a>
+        Absolutely positioned anchor
+      </h2>
+      <div style="position: relative" class="demo-elements">
```
jgerigmeyer, line 674:
```suggestion
      <div style="position: relative; height: 150px;" class="demo-elements">
```

index.html
```diff
@@ -663,6 +664,48 @@ <h2>
...
+        <div id="my-anchor-absolute-one" class="anchor">Anchor one</div>
+        <div id="my-target-absolute-one" class="target">Target one</div>
+        <div id="my-anchor-absolute-two" class="anchor">Anchor two</div>
+        <div id="my-target-absolute-two" class="target">Target two</div>
+      </div>
+      <p class="note">
+        With polyfill applied: Target and Anchor's right edges line up. Target's
+        top edge lines up with the bottom edge of the Anchor.
```
jgerigmeyer, line 682:
```suggestion
        With polyfill applied: Target one is positioned at the bottom right
        corner of Anchor one. Target two is positioned at the bottom right
        corner of Anchor two.
```

index.html
```diff
@@ -663,6 +664,48 @@ <h2>
...
+
+#my-anchor-absolute-two {
+  top: 120px;
+}
+
+#my-target-absolute-one,
+#my-target-absolute-two {
+  position: absolute;
+  top: anchor(--my-anchor-absolute bottom);
+  left: anchor(--my-anchor-absolute right);
```
jgerigmeyer, line 708:
```suggestion
>#my-anchor-absolute-one,
#my-anchor-absolute-two {
  position: absolute;
  anchor-name: --my-anchor-absolute;
}

#my-target-absolute-one,
#my-target-absolute-two {
  position: absolute;
  top: anchor(--my-anchor-absolute bottom);
  left: anchor(--my-anchor-absolute right);
```

public/anchor-absolute.css
```diff
@@ -0,0 +1,26 @@
+#anchor-absolute > .demo-elements {
+  height: 200px;
+}
+
+#my-anchor-absolute-one,
```
jgerigmeyer, line 5:
```suggestion
#my-anchor-absolute-one,
```

jgerigmeyer (commented):
Thanks for this contribution! Aside from the bug you and @jamesnw have already discussed, I made a few minor suggestions to simplify the example CSS.

index.html
```diff
@@ -697,6 +698,36 @@ <h2>
   top: anchor(--my-anchor-name-b bottom);
 }
 </code></pre>
+    </section>
+    <section id="anchor-absolute" class="demo-item">
+      <h2>
+        <a href="#anchor-absolute" aria-hidden="true">🔗</a>
+        Absolutely positioned anchor
+      </h2>
+      <div style="position: relative; height: 150px;" class="demo-elements">
```
jgerigmeyer, line 707: I guess the linter complains about this extra semicolon 🤷 

```suggestion
      <div style="position: relative; height: 150px" class="demo-elements">
```

jgerigmeyer (approved):
This is great -- thanks again! 👏 

> Some end-to-end tests that are outdated compared with spec started failing, should I rewrite them?

Yes, if you're willing to adjust the 5 failing tests, then this is good to merge.

jgerigmeyer:
@ayoreis We're going to merge this and fix up the tests on our end so we can cut a new release. Thanks again for your contribution!

ayoreis:
Thanks!

Issue #179, opened 2024-05-17
Default anchor: `position-anchor`


jamesnw:
Spec - https://drafts.csswg.org/css-anchor-position-1/#position-anchor

Issue #192, opened 2024-05-17
Anchors in some multicolumn scenarios positioned incorrectly


jamesnw:
Related WPT:

- anchor-position-inline-004.html
- anchor-position-multicol-002.html
- anchor-position-multicol-004.html
- anchor-position-multicol-colspan-001.html
- anchor-position-multicol-fixed-001.html
- anchor-position-multicol-nested-001.html

jamesnw:
May also be related to #188

Issue #180, opened 2024-05-17
Scoping anchor names: `anchor-scope`


jamesnw:
Spec: https://drafts.csswg.org/css-anchor-position/#anchor-scope

PR #196, opened 2024-05-23
Position anchor

jamesnw:
## Description
Implements the [`position-anchor`](https://drafts.csswg.org/css-anchor-position-1/#position-anchor) property.

## Related Issue(s)
#179 

## Steps to test/reproduce
View the [demo](https://deploy-preview-196--anchor-polyfill.netlify.app/#position-anchor) in a browser that needs the polyfill, and apply the polyfill.

## Show me
image

- [x] Update Tests
- [x] Confirm CSSTree parsing `.class#id` as `.class #id` is not an internal bug
- [x] Support `auto` value (defer until later)


src/parse.ts
```diff
@@ -394,12 +405,25 @@ function getAnchorFunctionData(
   return {};
 }
 
+function getPositionAnchorData(node: csstree.CssNode, rule?: csstree.Raw) {
+  if (isPositionAnchorDeclaration(node) && rule?.value) {
+    const targets = document.querySelectorAll(rule.value);
+    const name = getDeclarationValue(node);
+
+    for (const targetEl of targets) {
+      (targetEl as HTMLElement).dataset[POSITION_ANCHOR_DATASET_KEY] = name;
```
jamesnw, line 414: It seems odd to me that that I'm changing the DOM in a `get` function- would this be better outside of the function?

src/parse.ts
```diff
@@ -452,6 +476,8 @@ async function getAnchorEl(
       return await validatedForPositioning(targetEl, [
         `#${CSS.escape(anchorAttr)}`,
       ]);
+    } else if (targetEl.dataset[POSITION_ANCHOR_DATASET_KEY]) {
+      anchorName = targetEl.dataset[POSITION_ANCHOR_DATASET_KEY];
```
jamesnw, line 480: I tried putting the anchor name in a style attribute, but that meant I needed to parse the style attribute string, and the data attributes were cleaner.

I also considered just using the anchor attribute, but thought that could be confusing when a user is inspecting.

jamesnw: I considered removing the data attribute, but could only do so at the very end, when I'm sure there are no more elements that may need this. We could either do this with a `[data-polyfill-position-anchor-name]` query, or by storing the elements or selectors where we applied it. I'm not sure if it's more important to clean it up or to limit DOM interactions.

src/parse.ts
jgerigmeyer: Hm, I worried this won't be robust enough -- what happens with e.g.:

```html
<div id="anchored" />
```

```css
#anchored {
  position-anchor: --bar;
}
div {
  position-anchor: --foo;
}
```

I think we'll end up with `--foo` as the anchor for our div, where I think according to the cascade we'd want `--bar`?

What if we replace `position-anchor` (or add to it) in the CSS with a custom property, and then we look for the value of that custom property on the `targetEl` in `getAnchorEl`?

jgerigmeyer (changes requested):
@jamesnw I'm slowly trying to re-learn this codebase, so forgive me if my concern here is completely off! 😅

src/parse.ts
jamesnw: > What if we replace `position-anchor` (or add to it) in the CSS with a custom property, and then we look for the value of that custom property on the `targetEl` in `getAnchorEl`? That way we're grabbing the info we need, but letting the cascade do the work of applying the correct property to the correct elements.

Darn it, you're right that this isn't robust enough, and needs to take into account the cascade. I was easily able to break it with different CSS rules. 

I attempted to use a custom property, but I wasn't able to set a custom property and get the cascaded value in the same pass. 

1. In the line updating the CSS (`styleObj.css = csstree.generate(ast);`), I can see the custom property in the CSS.
2. Inside `getAnchorEl`, any changed CSS properties are removed completely from the element inspector in the dev tools, and `getComputedStyle(el)` does not contain the custom property, so I'm not able to access it
3. After the polyfill fully runs, I'm able to inspect the element, and the changed CSS properties are now present, and `getComputedStyle(el)` contains the CSS property.

Do we need to somehow allow the CSS to re-apply before `getAnchorEl` is called?

jamesnw: In looking into this a bit more, I think we want to add an initial pass to the polyfill, before the existing `parseCSS` step. We would limit this step to moving non-cascading values to custom properties (`position-anchor` to  `--position-anchor`, and likely `anchor-scope` to `--anchor-scope` will be needed for #180). 

We will need to adjust `transformCSS` to allow for multiple transformations of the same input styles, and will also want to do some benchmarking to ensure this doesn't cause too much of a performance hit.

The alternative would likely be ensuring that stylesheets are parsed in cascade order, and then calculating and comparing rule specificities to ensure the right rule wins.

jgerigmeyer: > In looking into this a bit more, I think we want to add an initial pass to the polyfill, before the existing `parseCSS` step. We would limit this step to moving non-cascading values to custom properties (`position-anchor` to `--position-anchor`, and likely `anchor-scope` to `--anchor-scope` will be needed for #180).
> 
> We will need to adjust `transformCSS` to allow for multiple transformations of the same input styles, and will also want to do some benchmarking to ensure this doesn't cause too much of a performance hit.

@jamesnw Yes, I think this ☝️ is correct. (Sorry it took me awhile to get back to this PR!)

jamesnw:
@jgerigmeyer  This is ready for re-review at your convenience.

I did opt to leave out support `position-anchor: auto`, which is "equivalent to omitting the anchor element in `anchor()`", as implicit anchors are still pending in the spec, it isn't tested yet in WPT, and it isn't implemented yet in Chrome.

jamesnw:
This does introduce a new regression in [anchor-position-multicol-colspan-001.html](https://wpt.fyi/results/css/css-anchor-position/anchor-position-multicol-colspan-001.html?label=master&label=experimental&aligned&q=css/css-anchor-position/anchor-position-multicol-colspan-001.html), `.target 3`.  This is related to #192.

This change passes 12 new subtests, mostly in areas where `position-anchor` is not the primary concept under test.

src/parse.ts
```diff
@@ -574,6 +587,17 @@ export async function parseCSS(styleData: StyleData[]) {
         },
       );
 
+      // Parse `position-anchor` data
+      const { name: positionAnchorName, selector: positionAnchorSelector } =
+        getPositionAnchorData(node, rule);
```
jamesnw, line 592: This is the second time we're finding the `position-anchor` declarations. The first time is in the cascade step. In theory, we could return the needed info from the cascade step, and pass it into `parse`. I opted not to do that, as we are checking on an existing walk through the styles, and not adding an additional walk.

src/cascade.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+  return {};
+}
+
+export async function cascadeCSS(styleData: StyleData[]) {
+  const changedStyles = [];
+  for (const styleObj of styleData) {
+    let changed = false;
+    const ast = getAST(styleObj.css);
+    csstree.walk(ast, function (node) {
```
jgerigmeyer, line 36: I think we could use `visit: 'Declaration'` here to speed things up a bit.

https://github.com/csstree/csstree/blob/master/docs/traversal.md#visit

src/parse.ts
jgerigmeyer: I'm confused why this entire block is needed at all. We've already added a `--position-anchor` property and are using that to grab the relevant anchor for a given target -- what is this pass doing? And isn't it backwards to be storing `positionAnchorName` as if its an `anchorName`?

src/parse.ts
```diff
@@ -574,6 +587,17 @@ export async function parseCSS(styleData: StyleData[]) {
...
+      if (positionAnchorName && positionAnchorSelector) {
+        if (anchorNames[positionAnchorName]) {
+          anchorNames[positionAnchorName].push(positionAnchorSelector);
+        } else {
+          anchorNames[positionAnchorName] = [positionAnchorSelector];
+        }
+      }
+
```
jgerigmeyer, line 600:
```suggestion
```

src/cascade.ts
```diff
@@ -0,0 +1,52 @@
...
+      type: 'Declaration',
+      important: false,
+      property: '--position-anchor',
+      value: {
+        type: 'Raw',
+        value: getDeclarationValue(node),
+      },
+    } as const;
+
+    block.children.append(block.children.createItem(newProp));
```
jgerigmeyer, line 25:
```suggestion
    block.children.appendData({
      type: 'Declaration',
      important: false,
      property: '--position-anchor',
      value: {
        type: 'Raw',
        value: getDeclarationValue(node),
      },
    });
```

src/cascade.ts
```diff
@@ -0,0 +1,52 @@
...
+      const { updated } = shiftPositionAnchorData(node, block);
+      if (updated) {
+        changed = true;
+      }
+    });
+    if (changed) {
+      // Update CSS
+      styleObj.css = generateCSS(ast);
+      styleObj.changed = true;
+      changedStyles.push(styleObj);
```
jgerigmeyer, line 48: If all we care about in the end is a boolean "did we change anything or not", it might be simpler just to return a boolean instead of an array that we won't use.

jgerigmeyer (commented):
This is looking great! A few remaining questions.

src/parse.ts
jamesnw: Good catch. Removed!

src/cascade.ts
jamesnw: Fixed in 7298e90e3d8ec4684fafee3537cafb3db1504a22

src/cascade.ts
jamesnw: Updated in 7298e90e3d8ec4684fafee3537cafb3db1504a22.

src/cascade.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Move `position-anchor` declaration to cascadable `--position-anchor`
+// property.
+function shiftPositionAnchorData(node: csstree.CssNode, block?: csstree.Block) {
+  if (isPositionAnchorDeclaration(node) && block) {
+    const newProp = {
+      type: 'Declaration',
+      important: false,
+      property: '--position-anchor',
```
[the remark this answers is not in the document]
jamesnw: Added in 7298e90e3d8ec4684fafee3537cafb3db1504a22

tests/unit/setup.ts
```diff
@@ -11,3 +11,6 @@ beforeAll(() => {
 afterEach(() => {
   fetchMock.reset();
 });
+vi.mock('nanoid/non-secure', () => {
+  return { nanoid: () => '1234' };
+});
```
jgerigmeyer: Is this needed? It seems like it's valuable for us to test using the real library unless there's a good reason not to, and the tests still pass for me after removing this.

jgerigmeyer: Removed in https://github.com/oddbird/css-anchor-positioning/commit/e205a2221d28f9a8b416596c9ceec310237ac8fe.

jgerigmeyer (approved):
🚀

tests/unit/setup.ts
jamesnw: It looks like I ended up always testing against the const exported from utils, so you're correct this isn't needed any longer.

jgerigmeyer:
@jamesnw I'm happy for you to merge this if you don't have more changes!

PR #202, opened 2024-06-14
Position Fallback

### LINKED ISSUE #167: Overflow management with position-try syntax
Hey 👋 

Been trying the polyfill with a few demos. Any plans to support the new syntax and auto flip values?

https://drafts.csswg.org/css-anchor-position/#position-try-options

jamesnw:
## Description
Changes from `@position-fallback` and `@try` to @position-try and related declarations.

Implements Position Fallback, with the following known exceptions:

  - `position-try-order`. If `try-size` is specified in `position-try` shorthand, it will be parsed, and `try-tactics` will be applied, but the`try-size` will be ignored.
  - The `flip-start` `try-tactic` is only partially supported. The tactic is only applied to property names and anchor sides.
  - a `position-area` as a `try-tactic`
  - Fallback does does not support anchor functions that are nested or passed through custom properties.
  -  Multiple selectors for a single `position-try-fallbacks` rule that uses an at position try rule is not supported.
  - Polyfill allows fallback in scroll more permissively than the spec allows, for instance without `position-anchor`.

## Related Issue(s)
Closes #167 

## Todo

- [x] Check Position-try example- not working in non-polyfilled Chromium
- [x] Check if there are issues with cascade, see https://github.com/oddbird/css-anchor-positioning/pull/202/files/b40520e6e37e7a03454c101fb5e434820d063fe8#r1681816222
- [x] Change `inset-area` to `position-area`


jamesnw:
Watch https://github.com/w3c/csswg-drafts/issues/10395 for potential renaming of `position-try-options` to `position-try-fallbacks`.

jamesnw:
> Watch [w3c/csswg-drafts#10395](https://github.com/w3c/csswg-drafts/issues/10395) for potential renaming of `position-try-options` to `position-try-fallbacks`.

Changed to match resolution in https://github.com/oddbird/css-anchor-positioning/pull/202/commits/786538b7d3f41d9d5f7058908c9e44860017290a

jamesnw:
Scroll triggering for fallbacks seems to be working different in Chromium than the Polyfill- I suspect it stems from https://github.com/w3c/csswg-drafts/issues/8059#issuecomment-1444616906, but more research is needed.

jamesnw:
@jgerigmeyer I think this is in a place for an initial review. It implements a subset of the total functionality, but it could be a useful release, with the TODOs broken out into separate issues.

I'm noting there are a lot of type definitions, consts, and assertions, with a few circular imports, and I'm wondering about your thoughts on pulling those out into a separate file, or set of files.

src/fallback.ts
```diff
@@ -0,0 +1,612 @@
...
+    if (isAnchorFunction(valueAst.children.first)) {
+      valueAst.children.first.children.map((item) => {
+        if (
+          item.type === 'Identifier' &&
+          ANCHOR_SIDES.includes(item.name as AnchorSideKeyword)
+        ) {
+          item.name = mapAnchorSide(item.name as AnchorSideKeyword, tactic);
+        }
+      });
+    }
```
jgerigmeyer, line 453: This feels pretty fragile, and seems like a simpler duplicate of the logic we're using elsewhere to parse anchor fns. What about one of these cases?

```css
#foo {
  position: absolute;
  --half: anchor(--my-anchor-props 50%);
  left: var(--half);
}
```

```css
#bar {
  position: absolute;
  left: calc(anchor(--my-anchor-math 100%) - 50px);
}
```

I'm not sure I have a good solution off the top of my head, but at the very least want to be sure we're aware of (and document) the known limitations. It'd be great if we could make use of the anchor-fn parsing we're already doing, since that is very carefully determining the relevant inset property and anchor side for every usage of `anchor()`, even if traced through custom properties.

If this ran _after_ that step, I wonder if we could check for known `var(--anchor-<uuid>)` values to match this usage up with our parsed data objects in `anchorFunctions`?

src/parse.ts
```diff
@@ -506,50 +494,106 @@ export async function parseCSS(styleData: StyleData[]) {
...
+          }
+          if (name && selector && fallbacks[name]) {
+            if (anchorPosition.fallbacks) {
+              anchorPosition.fallbacks = [
+                ...anchorPosition.fallbacks,
+                ...fallbacks[name].blocks,
+              ];
+            } else {
+              anchorPosition.fallbacks = fallbacks[name].blocks;
+            }
```
jgerigmeyer, line 546:
```suggestion
            anchorPosition.fallbacks ??= [];
            anchorPosition.fallbacks.push(...fallbacks[name].blocks);
```

src/parse.ts
```diff
@@ -506,50 +494,106 @@ export async function parseCSS(styleData: StyleData[]) {
...
+                blocks: [
+                  {
+                    uuid: `${tryObject.tactic}-try-${nanoid(12)}`,
+                    declarations: tacticAppliedRules,
+                  },
+                ],
+              };
+            }
+          }
+          if (name && selector && fallbacks[name]) {
```
jgerigmeyer, line 538:
```suggestion
          if (name && fallbacks[name]) {
```

src/parse.ts
```diff
@@ -506,50 +494,106 @@ export async function parseCSS(styleData: StyleData[]) {
...
+                  ),
+                },
+              });
+              // Store mapping of data-attr to target selector
+              fallbackTargets[dataAttr] = selector;
+            }
+            changed = true;
+          }
+        });
+        if (selector && Object.keys(anchorPosition).length > 0) {
```
jgerigmeyer, line 581:
```suggestion
        if (Object.keys(anchorPosition).length > 0) {
```

src/parse.ts
```diff
@@ -506,50 +494,106 @@ export async function parseCSS(styleData: StyleData[]) {
...
+          if (validPositions[selector]) {
+            if (anchorPosition.order) {
+              validPositions[selector].order = anchorPosition.order;
+            }
+            if (anchorPosition.fallbacks) {
+              if (!validPositions[selector].fallbacks)
+                validPositions[selector].fallbacks = [];
+              validPositions[selector].fallbacks?.push(
+                ...anchorPosition.fallbacks,
+              );
```
jgerigmeyer, line 591:
```suggestion
              validPositions[selector].fallbacks ??= [];
              validPositions[selector].fallbacks.push(
                ...anchorPosition.fallbacks,
              );
```

src/fallback.ts
```diff
@@ -0,0 +1,612 @@
...
+      valueAst.children.map((id) => {
+        if (id.type === 'Identifier' && isInsetAreaProp(id.name)) {
+          id.name = mapInsetArea(id.name, tactic);
+        }
+        return id;
+      });
+    }
+    if (key.startsWith('margin')) {
+      mapMargin(key, valueAst, tactic);
+    }
```
jgerigmeyer, line 465: I admittedly didn't review the logic for these very carefully 😅

jgerigmeyer (commented):
@jamesnw Whew! This is a lot, and good progress! Most of my comments are very minor, or questions to be sure I understand what's going on. I think the main substantive concern I have is around how we determine the proper values to "flip" in a robust way -- essentially the combination of `shiftInsetProperties` and `applyTryTactic`, which you identified early on as the main difficulty here. We're already doing a lot of work to carefully track `anchor()` usage down to a particular anchor-side/percentage and inset-property, and ideally we could re-use that work for the `position-try-fallbacks` logic as well.

Happy to jump on a call to talk through this -- it may be easier to rubber-duck it together.

jgerigmeyer:
> I'm noting there are a lot of type definitions, consts, and assertions, with a few circular imports, and I'm wondering about your thoughts on pulling those out into a separate file, or set of files.

Yes, this is 💯 -- maybe we want to file an issue specifically for file sorting and function documentation?

index.html
```diff
@@ -469,40 +469,102 @@ <h2>
...
+          <li>
+            When every position overflows, revert to the initial base styles in
+            #1, even though it overflows.
+          </li>
+        </ol>
+        <p>
+          <strong>Note:</strong> Early Chromium implementations used
+          <code>position-try-options</code> instead of
+          <code>position-try-fallbacks</code>. We recommend adding both
+          properties, or using the <code>position-try</code> shorthand.
```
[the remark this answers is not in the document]
jamesnw: The polyfill does support `position-try`- it parses out the options, and uses the values when it can. So `position-try: most-block-size flip-block, flip-inline` would use the `flip-block` and `flip-inline`, but it doesn't yet do anything with `most-block-size`.

src/fallback.ts
```diff
@@ -0,0 +1,612 @@
...
+      declarations[key] ??= 'revert';
+    }
+
+    if (isAnchorFunction(valueAst.children.first)) {
+      valueAst.children.first.children.map((item) => {
+        if (
+          item.type === 'Identifier' &&
+          ANCHOR_SIDES.includes(item.name as AnchorSideKeyword)
+        ) {
+          item.name = mapAnchorSide(item.name as AnchorSideKeyword, tactic);
```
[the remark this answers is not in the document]
jamesnw: I'm actually editing the ast children object by reference, so I switched to use `children.forEach`, and not return the child value.

src/cascade.ts
```diff
@@ -32,22 +39,36 @@ function shiftUnsupportedProperties(
   return {};
 }
 
-/**
- * Update the given style data to enable cascading and inheritance of properties
- * that are not yet natively supported.
- */
-export function cascadeCSS(styleData: StyleData[]) {
+// Move all inset properties to custom properties
+function shiftInsetProperties(
```
mmalerba, line 43: You should just be able to add mapping for these to `SHIFTED_PROPERTIES` above. That also has the advantage of working with plain old `getCSSPropertyValue(el, 'left')`

src/utils.ts
```diff
@@ -35,3 +44,26 @@ export interface StyleData {
...
+        acc[acc.length - 1].push(child);
+      }
+
+      return acc;
+    },
+    [[]],
+  );
+}
+
+export function getCSSPropertyValue(el: HTMLElement, prop: string) {
```
mmalerba, line 67: I made some tweaks to this function and moved it to src/dom.ts

src/polyfill.ts
```diff
@@ -349,38 +374,30 @@ async function applyPositionFallbacks(
           return;
         }
         checking = true;
+        target.removeAttribute('data-anchor-polyfill');
+        const defaultOverflow = await checkOverflow(target, offsetParent);
```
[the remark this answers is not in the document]
jamesnw: My guess it is isn't doing any intelligent batching here- I split this into a new issue #227 to revisit this.

src/fallback.ts
```diff
@@ -0,0 +1,612 @@
...
+function parsePositionTryFallbacks(list: csstree.List<csstree.CssNode>) {
+  const positionOptions = splitCommaList(list);
+  const tryObjects: PositionTryObject[] = [];
+  positionOptions.forEach((option) => {
+    if (option.length === 2 && isPositionTryTactic(option[0].name)) {
+      tryObjects.push({
+        tactic: option[0].name,
+        atRule: option[1].name,
+        type: 'at-rule-with-try-tactic',
+      });
```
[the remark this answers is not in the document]
jamesnw: This now supports either order.

README.md
```diff
@@ -82,7 +82,7 @@ While this polyfill supports many basic use cases, it doesn't (yet) support the
 following features:
 
 - The `@position-try` rule
-- The `position-try-options`, `position-try-order`, or `position-try` properties
+- The `position-try-fallbacks`, `position-try-order`, or `position-try` properties
```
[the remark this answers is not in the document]
jamesnw: Updated in https://github.com/oddbird/css-anchor-positioning/pull/202/commits/ada0768fb30b0080d1ac4047bb23607d60ca0edd

index.html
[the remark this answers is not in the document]
jamesnw: Updated in https://github.com/oddbird/css-anchor-positioning/pull/202/commits/ada0768fb30b0080d1ac4047bb23607d60ca0edd

src/parse.ts
```diff
@@ -48,13 +48,46 @@ export const INSET_PROPS: InsetProperty[] = [
...
+  | 'margin-inline-start'
+  | 'margin-inline-end'
+  | 'margin-inline'
+  | 'margin-bottom'
+  | 'margin-left'
+  | 'margin-right'
+  | 'margin-top'
+  | 'margin';
+
+export const MARGIN_PROPERTIES: MarginProperty[] = [
```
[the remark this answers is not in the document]
jamesnw: Thanks, I transitioned to this style in https://github.com/oddbird/css-anchor-positioning/pull/202/commits/aaa4f66ad5353a356c7291f1fa4f305ab5f56201

src/fallback.ts
```diff
@@ -0,0 +1,740 @@
...
+            const tacticAppliedRules = applyTryTacticsToSelector(
+              selector,
+              tryObject.tactics,
+            );
+            if (tacticAppliedRules) {
+              fallbacks[name] = {
+                targets: [selector],
+                blocks: [
+                  {
+                    uuid: `${tryObject.tactics.join('-')}-try-${nanoid(12)}`,
```
jgerigmeyer, line 655: Do we want to prepend `${selector}-` here?

src/fallback.ts
```diff
@@ -0,0 +1,740 @@
...
+    let changed = false;
+    const ast = getAST(styleObj.css);
+    csstree.walk(ast, {
+      visit: 'Declaration',
+      enter(node) {
+        const rule = this.rule?.prelude as csstree.SelectorList | undefined;
+        const selectors = getSelectors(rule);
+        if (!selectors.length) return;
+        // todo: better handle multiple selectors
+        const selector = selectors.map((s) => s.selector).join(',');
```
jgerigmeyer, line 629: We're only using `selector` to pass through to `applyTryTacticsToSelector` -- should we just loop through `selectors` in those cases, calling `applyTryTacticsToSelector` for each selector individually?

src/fallback.ts
```diff
@@ -0,0 +1,740 @@
...
+            }
+          }
+
+          if (name && fallbacks[name]) {
+            anchorPosition.fallbacks ??= [];
+            anchorPosition.fallbacks.push(...fallbacks[name].blocks);
+
+            if (!fallbacks[name].targets.includes(selector)) {
+              fallbacks[name].targets.push(selector);
+            }
```
jgerigmeyer, line 689: Should we do this `for (const { selector } of selectors)`?

src/fallback.ts
```diff
@@ -0,0 +1,745 @@
...
+      declarations[key] ??= 'revert';
+    }
+
+    if (isAnchorFunction(valueAst.children.first)) {
+      valueAst.children.first.children.forEach((item) => {
+        if (isIdentifier(item) && isAnchorSide(item.name)) {
+          item.name = mapAnchorSide(item.name, tactic);
+        }
+      });
+    }
```
jgerigmeyer, line 433: I'm pretty sure this still has the same limitations outlined in https://github.com/oddbird/css-anchor-positioning/pull/202#discussion_r1681816222 -- is that right?

I'm okay with that for now, but we should add a detailed comment here explaining that nested `anchor()` fns (or passed through custom properties) won't work with `try-tactics`.

jgerigmeyer (commented):
Whew! Again, a lot here -- the revised spec really made our jobs harder 😅 

This is obviously an improvement over the current entire lack of support!

src/fallback.ts
jamesnw: Adjusted in fd96e3068816d185ec5fcde2afcfd606b717aaf5

src/fallback.ts
jamesnw: Changed in fd96e3068816d185ec5fcde2afcfd606b717aaf5

jamesnw (commented):
Thanks @jgerigmeyer! I've made the changes, and documented in the PR the limitations. I'll open issues once this gets merged.

src/syntax.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Accepted position try properties
+export const ACCEPTED_POSITION_TRY_PROPERTIES = [
+  ...INSET_PROPS,
+  ...MARGIN_PROPERTIES,
+  ...SIZING_PROPS,
+  ...SELF_ALIGNMENT_PROPS,
+  'position-anchor',
+  'inset-area',
```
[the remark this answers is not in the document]
jamesnw: Updated to `position-area`.

src/fallback.ts
```diff
@@ -0,0 +1,755 @@
...
+      enter(node) {
+        // Parse `@position-try` rules
+        const { name, blocks } = getPositionTryRules(node);
+        if (name && blocks?.length) {
+          // This will override earlier `@position-try` lists with the same
+          // name: (e.g. multiple `@position-try --my-fallback {...}` uses with
+          // the same `--my-fallback` name)
+
+          // Todo: this doesn't account for multiple selectors for a single
+          // `position-try-fallbacks` rule that uses an `@position-try` rule.
```
jgerigmeyer, line 616: I'm not sure I understand this comment -- is it out of date?

src/fallback.ts
```diff
@@ -0,0 +1,755 @@
...
+      else if (opt.name.startsWith('--')) identifiers.atRules.push(opt.name);
+      else if (isPositionAreaProp(opt.name))
+        identifiers.positionAreas.push(opt.name);
+    });
+    // Position area can not be combined or have multiple
+    if (identifiers.positionAreas.length) {
+      tryObjects.push({
+        positionArea: identifiers.positionAreas[0],
+        type: 'position-area',
+      });
```
jgerigmeyer, line 478: I switched this in https://github.com/oddbird/css-anchor-positioning/commit/c6959899fcda8b6444dbfcfa67c540d139642565 to use `isPositionAreaProp` instead of `isInsetProp` -- I think that's correct?

jgerigmeyer (approved):
Looks good!

Todo:
- [ ] open issues for todos
- [ ] add todos to Limitations section of readme

src/fallback.ts
jamesnw: Good catch! That is correct.

src/fallback.ts
jamesnw: No, this is still the case- Something like this will not work for both selectors. 

```css
#targetA, #targetB{
  position-try-fallbacks: --my-at-rule;
}
```

Only 1 fallback will be made, but the at rule will apply to  `#targetA` and `#targetB` differently.

jgerigmeyer: Hm, I'm still not following. Only one `@position-try` definition will be stored (which is correct), but then the `selectors.forEach(...` loop will run for each selector, and the same `@position-try` definition will be stored in `validPositions[selector]`. I would think that should work? Let's test it out.

jgerigmeyer: I think this is resolved in #235

PR #206, opened 2024-06-26
Remove unneeded Polyfill

jamesnw:
## Description
Removes CDN import from polyfill.io, which is no longer recommended.

I don't think we need a Promise Polyfill anymore, but if we do, we have options from https://community.fastly.com/t/new-options-for-polyfill-io-users/2540.

### DIFF
--- a/tests/runner.html
+++ b/tests/runner.html
@@ -1,8 +1,5 @@
 <!doctype html>
 <html>
-  <head>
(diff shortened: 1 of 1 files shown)


jgerigmeyer (approved):

jgerigmeyer:
This was only used by the WPT runner, and that's already broken at the moment.

PR #213, opened 2024-07-02
Pseudo-element anchors

### LINKED ISSUE #208: Anchoring on pseudo elements
Pseudo elements should be supported as valid anchors. 

From the [spec](https://drafts.csswg.org/css-anchor-position-1/#target), a  `part-like pseudo-element` is a valid anchor.

[Implicit](https://drafts.csswg.org/css-anchor-position-1/#implicit) anchoring of pseudo-elements is specified as well.

ayoreis:
TODO:
- [x] Fix the "Positioning with anchor() [passed through multiple CSS custom properties]" example. `customPropsMapping` seems not to get populated.
- [x] Update tests to expect pseudo elements.
- [ ] Get WPT passing for pseudo-element anchors. The polyfill is working (if I manually check the conditions, eg. `anchored1.offsetTop === 100`) but WPT reports `0`, no tests went down.

~Performance does not seem worse (tested with `console.time` consistently at \~110ms).~

Closes #208.

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -9,15 +9,20 @@ import {
   POSITION_ANCHOR_PROPERTY,
   type StyleData,
 } from './utils.js';
-import { validatedForPositioning } from './validate.js';
+import { type PseudoElement, validatedForPositioning } from './validate.js';
 
 interface AtRuleRaw extends csstree.Atrule {
   prelude: csstree.Raw | null;
 }
 
+export interface AnchorSelector {
+  selector: string;
+  pseudoElement?: string;
+}
+
 // `key` is the `anchor-name` value
 // `value` is an array of all element selectors with that anchor name
-type AnchorNames = Record<string, string[]>;
+type AnchorNames = Record<string, AnchorSelector[]>;
 
 export type InsetProperty =
   | 'top'
@@ -124,7 +129,7 @@ const ANCHOR_SIZES: AnchorSize[] = [
 
 export interface AnchorFunction {
   targetEl?: HTMLElement | null;
-  anchorEl?: HTMLElement | null;
+  anchorEl?: HTMLElement | PseudoElement | null;
   anchorName?: string;
   anchorSide?: AnchorSide;
   anchorSize?: AnchorSize;
@@ -341,18 +346,31 @@ function parseAnchorFn(
   };
 }
 
-function getAnchorNameData(node: csstree.CssNode, rule?: csstree.Raw) {
-  if (
-    isAnchorNameDeclaration(node) &&
-    !node.value.children.isEmpty &&
-    rule?.value
-  ) {
-    return node.value.children.map((item) => {
-      const { name } = item as csstree.Identifier;
-      return { name, selector: rule.value };
-    });
-  }
-  return [];
+function getAnchorNames(node: DeclarationWithValue) {
+  return (node.value.children as csstree.List<csstree.Identifier>).map(
+    ({ name }) => name,
+  );
+}
+
+function getAnchorSelectors(selectors: csstree.SelectorList) {
+  return (selectors.children as csstree.List<csstree.Selector>).map(
+    (selector) => {
+      let pseudoElement: string | undefined;
+
+      if (selector.children.last?.type === 'PseudoElementSelector') {
+        selector = csstree.clone(selector) as csstree.Selector;
+        pseudoElement = (
+          selector.children.last as csstree.PseudoElementSelector
+        ).name;
+        selector.children.pop();
+      }
+
+      return {
+        selector: generateCSS(selector),
+        pseudoElement,
+      };
+    },
+  );
 }
 
 let anchorNames: AnchorNames = {};
@@ -468,7 +486,7 @@ async function getAnchorEl(
       anchorName = getCSSPropertyValue(targetEl, customPropName);
     } else if (anchorAttr) {
       return await validatedForPositioning(targetEl, [
-        `#${CSS.escape(anchorAttr)}`,
+        { selector: `#${CSS.escape(anchorAttr)}` },
       ]);
     }
   }
@@ -564,21 +582,21 @@ export async function parseCSS(styleData: StyleData[]) {
     let changed = false;
     const ast = getAST(styleObj.css);
     csstree.walk(ast, function (node) {
-      const rule = this.rule?.prelude as csstree.Raw | undefined;
+      const selectorList = this.rule?.prelude as
+        | csstree.SelectorList
+        | undefined;
+
+      const anchorSelectors = selectorList
+        ? getAnchorSelectors(selectorList)
+        : undefined;
 
       // Parse `anchor-name` declaration
-      const anchorNameData = getAnchorNameData(node, rule);
-      anchorNameData.forEach(
-        ({ name: anchorName, selector: anchorSelector }) => {
-          if (anchorName && anchorSelector) {
-            if (anchorNames[anchorName]) {
-              anchorNames[anchorName].push(anchorSelector);
-            } else {
-              anchorNames[anchorName] = [anchorSelector];
-            }
-          }
-        },
-      );
+      if (isAnchorNameDeclaration(node) && anchorSelectors) {
+        for (const name of getAnchorNames(node)) {
+          anchorNames[name] ??= [];
+          anchorNames[name].push(...anchorSelectors);
+        }
+      }
 
       // Parse `anchor()` function
       const {
@@ -586,16 +604,22 @@ export async function parseCSS(styleData: StyleData[]) {
         data,
         changed: updated,
       } = getAnchorFunctionData(node, this.declaration);
-      if (prop && data && rule?.value) {
+      if (prop && data && anchorSelectors) {
         // This will override earlier declarations
         // with the same exact rule selector
         // *and* the same exact declaration property:
         // (e.g. multiple `top: anchor(...)` declarations
         // for the same `.foo {...}` selector)
-        anchorFunctions[rule.value] = {
-          ...anchorFunctions[rule.value],
-          [prop]: [...(anchorFunctions[rule.value]?.[prop] ?? []), data],
-        };
+        for (const { selector, pseudoElement } of anchorSelectors) {
+          const fullSelector = pseudoElement
+            ? `${selector}::${pseudoElement}`
+            : selector;
+
+          anchorFunctions[selector] = {
+            ...anchorFunctions[fullSelector],
+            [prop]: [...(anchorFunctions[fullSelector]?.[prop] ?? []), data],
+          };
+        }
       }
       if (updated) {
         changed = true;
@@ -718,11 +742,15 @@ export async function parseCSS(styleData: StyleData[]) {
     csstree.walk(ast, {
       visit: 'Function',
       enter(node) {
-        const rule = this.rule?.prelude as csstree.Raw | undefined;
+        const selectorList = this.rule?.prelude as
+          | csstree.SelectorList
+          | undefined;
+
         const declaration = this.declaration;
         const prop = declaration?.property;
+
         if (
-          rule?.value &&
+          selectorList?.children &&
           isVarFunction(node) &&
           declaration &&
           prop &&
@@ -743,7 +771,6 @@ export async function parseCSS(styleData: StyleData[]) {
           }
 
           /*
-
             An anchor (or anchor-size) fn was assigned to an inset (or sizing)
             property.
 
@@ -785,8 +812,7 @@ export async function parseCSS(styleData: StyleData[]) {
               #target {
                 top: var(--center-top-EnmDEkZ5mBLp);
               }
-
-            */
+          */
           const propUuid = `${prop}-${nanoid(12)}`;
 
           // If this is a custom property which was assigned a value from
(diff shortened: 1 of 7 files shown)


ayoreis:
I updated the tests and fixed `@position-fallback`. Also now instead of inserting the pseudo-element's `content` as text into the fake pseudo-element, it add it to the `::before` of the fake pseudo-element supporting all types (eg. images).

Currently all these changes add about 15ms to the polyfill (on the homepage, on my machine). It seem that the increase is only because of enabling `parseRulePrelude` for csstree, so it won't increase much with the amount of pseudo-element anchors, so that amount seems fine.

I looked into WPT, the problem is that `--inject-script` is sync and we have async code, so WPT runs before everything has been polyfilled, I think `window.CHECK_LAYOUT_DELAY` in `src/index-wpt.ts` should fix this but I can't find any docs on it to continue debugging.

One unrelated WPT test started passing with these changes (`/css/css-anchor-position/anchor-query-custom-property-registration.html`).

jamesnw:
> I looked into WPT, the problem is that `--inject-script` is sync and we have async code, so WPT runs before everything has been polyfilled, I think `window.CHECK_LAYOUT_DELAY` in `src/index-wpt.ts` should fix this but I can't find any docs on it to continue debugging.


`window.CHECK_LAYOUT_DELAY` was added to the WPT in order to make this polyfill testable. It is currently used in [checkLayoutForAnchorPos](https://github.com/web-platform-tests/wpt/blob/master/css/css-anchor-position/support/test-common.js#L36) to add a delay of 3 animation frames before checking layout. See [this comment ](https://issues.chromium.org/issues/40059176#comment149)for more info. 

The pseudoelement tests are not using `checkLayoutForAnchorPos`, and so I think we'll need another strategy. Perhaps we add a function to WPT that we can use to delay other tests that aren't using `checkLayoutForAnchorPos`? I'm guessing this is also the case for other test files as well.

ayoreis:
I'm not familiar with the WPT codebase, but could `<script type="module">` injected scripts be implemented? If yes, I might be interested in working on it with some guidance.

jamesnw:
> I'm not familiar with the WPT codebase, but could `<script type="module">` injected scripts be implemented? If yes, I might be interested in working on it with some guidance.

Yeah- there are some `<script type="module">` examples on other anchor position tests. I think you can run `./wpt serve --inject-script=[polyfillpath]` and run it locally (note `serve` rather than `run` in this context. You can navigate to http://web-platform.test:8000/css/css-anchor-position/pseudo-element-anchor.html for example and that should have the polyfill loaded.

ayoreis:
I think I didn't explain well. What I meant is: Is it possible to implement support for injecting `type="module"` scripts that run before the test, for example this is the resulting test with `--inject-script`:

```html
<!DOCTYPE html>
<script>
await console.log('Hello') // throws

// Remove the injected script tag from the DOM.
document.currentScript.remove();
</script>

<title>Test</title>

<div id="element">Test</div>

<script>
  test(() => {
    assert_equals(element.textContent, 'Test');
  }, 'Test');
</script>
```

And with something like `--inject-script-module`, if it could be implemented:

```html
<!DOCTYPE html>
<title>Test</title>

<div id="element">Test</div>

<script type="module">
await console.log('Hello') 

// Remove the injected script tag from the DOM.
document.currentScript.remove();
</script>

<script>
  test(() => {
    assert_equals(element.textContent, 'Test');
  }, 'Test');
</script>
```

This would solve our use case.

A issue with this idea is that all tests would have to be converted to `<script type="module">`s (because `type="module"` scripts defer), a workaround could be for the `test` function to wrap the callback in an `addEventListener('load')`, but I don't know if that would cover all cases.

jamesnw:
> I think I didn't explain well.

No- you explained well, on my second reading I realized I had misunderstood, but you got your response up first.

To boil down the problem, the tests are running before the polyfill is complete. Ideally, we would add some delays for testing the polyfill that wouldn't impact non-polyfill tests. I don't think that adding an injected module script would be simple. 

We do have a few tools at our disposal- `pseudo-element-anchor.html` works if you use `promise_test` and `waitForAtLeastOneFrame()`-

```
<script>
  promise_test(async () => {
    await waitForAtLeastOneFrame();
    assert_equals(anchored1.offsetLeft, 100);
    assert_equals(anchored1.offsetTop, 100);
  }, "::before as anchor");
  promise_test(async () => {
    await waitForAtLeastOneFrame();
    assert_equals(anchored2.offsetLeft, 100);
    assert_equals(anchored2.offsetTop, 100);
  }, "::after as anchor");
</script>
```

But that also impacts non-polyfilled tests. @jgerigmeyer  Do you have any thoughts on an analog to the `window.checkLayoutForAnchorPos` might be? I'm wondering about a `waitForAtLeastOneFrame` wrapper that only waits if `CHECK_LAYOUT_DELAY` is true?

jgerigmeyer:
> @jgerigmeyer Do you have any thoughts on an analog to the `window.checkLayoutForAnchorPos` might be? I'm wondering about a `waitForAtLeastOneFrame` wrapper that only waits if `CHECK_LAYOUT_DELAY` is true?

@jamesnw Yeah, when we landed https://github.com/web-platform-tests/wpt/pull/38442 we knew that there were a few WPTs that called `test()` directly, and it didn't fix those cases. We might want to check to mfreed to see what the WPT folks would be open to, but I could see adding something similar to `checkLayoutForAnchorPos` that would replace `test()` for anchor-position tests only.

jamesnw:
I opened a PR on oddbird:wpt https://github.com/oddbird/wpt/pull/3/files that adds a test function which waits for the polyfill to run. It makes `pseudo-element-anchor.html` pass, but doesn't work on `pseudo-element-anchor-dynamic.html` because dynamic changes aren't yet supported.

src/validate.ts
```diff
@@ -1,6 +1,89 @@
...
+
+    if (!pseudoElementPart) {
+      result.push(...elements);
+      continue;
+    }
+
+    for (const element of elements) {
+      const computedStyle = getComputedStyle(element, pseudoElementPart);
+      const fakePseudoElement = document.createElement('div');
+      const styleSheet = new CSSStyleSheet();
```
jamesnw, line 33: `CSSStyleSheet()` and `adoptedStyleSheets` are only [Baseline 2023](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet). 

@jgerigmeyer  Can we use these, or do we need to use a <style> element? Regrettably, we can't use the style attributes to target the pseudo elements.

jamesnw (commented):
I think this is looking pretty good! 

At some point, could you also open issues for the things we are not covering here- other pseudo elements like the file upload input button, or scrolling?

src/parse.ts
```diff
@@ -341,18 +347,34 @@ function parseAnchorFn(
...
+}
+
+function getSelectors(rule: csstree.SelectorList | undefined) {
+  if (!rule) return [];
+
+  return (rule.children as csstree.List<csstree.Selector>)
+    .map((selector) => {
+      let pseudoElementPart: string | undefined;
+
+      if (selector.children.last?.type === 'PseudoElementSelector') {
```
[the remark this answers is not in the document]
ayoreis: It seems no browser implements pseudo-classing pseudo-elements yet (tested on the latest Firefox, Chromium and GNOME Web (WebKit)), or at least I couldn't get it to work:

```html
<style>
    div::before {
        content: '::before';
        background-color: blue;
    }

    div::before:hover {
        background-color: red;
    }
</style>

<div>Hello world</div>
```

So this should not be a problem.

The Selectors Level 4 spec does define it though:
https://www.w3.org/TR/selectors-4/#pseudo-element-states
https://drafts.csswg.org/selectors/#pseudo-element-states

src/validate.ts
```diff
@@ -1,6 +1,89 @@
...
+      const startingScrollX = window.scrollX;
+      const boundingClientRect = fakePseudoElement.getBoundingClientRect();
+
+      result.push({
+        fakePseudoElement,
+        computedStyle,
+
+        removeFakePseudoElement() {
+          fakePseudoElement.remove();
+          document.adoptedStyleSheets = originalStyles;
```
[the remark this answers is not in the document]
ayoreis: It shouldn't because `removeFakePseudoElement` is called before the next one is created (in `validatedForPositioning`), I tested it on a website with multiple pseudo-element anchors and everything worked.

src/validate.ts
ayoreis: Didn't notice it was so new, `<style>` elements should work.

src/parse.ts
```diff
@@ -718,11 +750,12 @@ export async function parseCSS(styleData: StyleData[]) {
...
       visit: 'Function',
       enter(node) {
-        const rule = this.rule?.prelude as csstree.Raw | undefined;
+        const rule = this.rule?.prelude as csstree.SelectorList | undefined;
         const declaration = this.declaration;
         const prop = declaration?.property;
+
         if (
-          rule?.value &&
+          rule?.children &&
```
jgerigmeyer, line 758:
```suggestion
          rule?.children.isEmpty === false &&
```

src/parse.ts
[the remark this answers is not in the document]
jamesnw: Good point- I also don't see any tests in WPT. The best related issue I could find is [MDN](https://github.com/mdn/content/issues/32971)  having similar questions about the difference in support and spec. I don't think we need to worry about `::before:hover` at this point.

src/validate.ts
```diff
@@ -1,6 +1,89 @@
...
-import { getCSSPropertyValue } from './parse.js';
+import { getCSSPropertyValue, type Selector } from './parse.js';
+
+export interface PseudoElement extends VirtualElement {
+  fakePseudoElement: HTMLElement;
+  computedStyle: CSSStyleDeclaration;
+  removeFakePseudoElement(): void;
+}
+
+function getAnchorsBySelectors(selectors: Selector[]) {
```
jgerigmeyer, line 12: I think this fn would benefit from a few inline comments narrating what's being done.

src/validate.ts
jgerigmeyer: This similar section from the Popover polyfill might be a useful reference: https://github.com/oddbird/popover-polyfill/blob/main/src/popover.ts#L120-L142

src/validate.ts
```diff
@@ -1,6 +1,89 @@
...
+
+      fakePseudoElement.id = `fake-pseudo-element-${nanoid()}`;
+
+      for (const property of Array.from(computedStyle)) {
+        const value = computedStyle.getPropertyValue(property);
+        fakePseudoElement.style.setProperty(property, value);
+      }
+
+      styleSheet.insertRule(
+        `#${fakePseudoElement.id}::before { content: ${computedStyle.content}; }`,
```
jgerigmeyer, line 44: Should this always be `::before`, or should it match `pseudoElementPart`?

jgerigmeyer (commented):
A few questions and minor suggestions, but this is looking good!

src/validate.ts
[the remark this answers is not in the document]
ayoreis: Good point, I'll find an alternative for `CSSStyleSheets` or only use `<style>` elements.

src/validate.ts
ayoreis: Is there any advantage to using `CSSStyleSheet`s (when supported) to only using `<style>` elements?

jgerigmeyer: Not that I'm aware of... [This note](https://github.com/oddbird/popover-polyfill/blob/main/README.md?plain=1#L128-L131) implies that injecting `<style>` elements may impact mutation observers 🤷

src/validate.ts
ayoreis: I don't think it makes a difference, but I'll update it just to make sure.

src/parse.ts
```diff
@@ -514,14 +542,19 @@ export async function parseCSS(styleData: StyleData[]) {
...
+        const rule = this.rule?.prelude as csstree.SelectorList | undefined;
+        const selectors = getSelectors(rule);
+
         // Parse `position-fallback` declaration
-        const { name, selector } = getPositionFallbackDeclaration(node, rule);
-        if (name && selector && fallbacks[name]) {
-          validPositions[selector] = { fallbacks: fallbacks[name].blocks };
-          if (!fallbacks[name].targets.includes(selector)) {
-            fallbacks[name].targets.push(selector);
+        const { name } = getPositionFallbackDeclaration(node, rule);
```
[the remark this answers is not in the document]
ayoreis: I was already like that so I don't know why, sam in the old `getAnchorNameData`.

src/parse.ts
```diff
@@ -547,7 +580,9 @@ export async function parseCSS(styleData: StyleData[]) {
               },
             });
             // Store mapping of data-attr to target selector
-            fallbackTargets[dataAttr] = selector;
+            for (const { selector } of selectors) {
+              fallbackTargets[dataAttr] = selector;
```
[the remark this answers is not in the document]
ayoreis: Ups, don't what was I thinking. It should be a `.join` of all of them.

jgerigmeyer (approved):
This looks good to me! 🚀

jgerigmeyer:
@ayoreis I'm going to go ahead and merge this -- thanks!

I think the only outstanding somewhat-related item I see is this, but it's not urgent/blocking:

> At some point, could you also open issues for the things we are not covering here- other pseudo elements like the file upload input button, or scrolling?

PR #217, opened 2024-07-20
Add `anchor-scope` support

mmalerba:


jgerigmeyer (changes requested):
@mmalerba This is great -- thanks! 🚀 A few tasks before it's ready to merge:

- [ ] There are some tests in `validate.test.ts` that are using the outdated call signature for `validatedForPositioning` and `isAcceptableAnchorElement`.
- [ ] We merged in #213 this morning, and there are a few merge conflicts -- could you rebase off of `main`?

src/parse.ts
```diff
@@ -252,10 +265,12 @@ export function isBoxAlignmentProp(
...
 
-export function isPositionAnchorDeclaration(
-  node: csstree.CssNode,
-): node is DeclarationWithValue {
-  return node.type === 'Declaration' && node.property === 'position-anchor';
+export function isValidAnchorScopeValue(value: string) {
+  return (
+    value.startsWith('--') ||
+    value === AnchorScopeValue.All ||
+    value == AnchorScopeValue.None
```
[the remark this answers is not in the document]
mmalerba: Fixed the tests, rebased, and pushed. I actually removed this validation when I was rebasing, because I noticed that `anchor-name` wasn't really doing it anyways, and I think it would make more sense to just strip invalid declarations in the pre-processing (cascade) step. Then we can just assume things are valid in the later phases

src/parse.ts
```diff
@@ -258,10 +271,12 @@ export function isBoxAlignmentProp(
...
+  );
 }
```
jgerigmeyer, line 281:
```suggestion
```

src/validate.ts
```diff
@@ -270,14 +322,19 @@ export async function validatedForPositioning(
 
   const anchorElements = getAnchorsBySelectors(anchorSelectors);
 
+  // TODO: handle anchor-scope for pseudo-elements.
```
jgerigmeyer, line 325: If we're not planning to handle this in this PR, could you file a new issue for it?

jgerigmeyer (approved):
👍 

I think you could remove `anchor-scope` from the "Limitations" list in the README, or add a clarification that it's not supported for pseudo-elements.

PR #221, opened 2024-07-23
Fix pseudo-element anchors with `overflow: scroll;` parents

ayoreis:
I started to write the issue but then it felt simpler to fix it :)

The only thing I'm not sure about is the demo, I changed it for this case but now there isn't one without scrolling, having both maybe gets too long?

### DIFF
--- a/index.html
+++ b/index.html
@@ -777,7 +777,13 @@
         Pseudo-element anchor
       </h2>
       <div style="position: relative" class="demo-elements">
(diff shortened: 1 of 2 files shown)


jgerigmeyer (approved):
> The only thing I'm not sure about is the demo, I changed it for this case but now there isn't one without scrolling, having both maybe gets too long?

I'm fine with this as-is 👍

Issue #212, opened 2024-07-02
[BUG] Media Queries are not applied to `anchor-name`


jamesnw:
**Describe the bug**
If a `anchor-name` is defined inside a media query, the context of the media query is stripped out. This means that you can't use the polyfill to dynamically select the anchor based on screen size or other media queries.

**To Reproduce**
1. Go to https://codepen.io/jamessw/pen/XWwveBJ in a browser that requires the polyfill
2. See that the target aligns with the anchor that is defined in a `@media print` block, even though the non-polyfilled declarations are not applied.

**Expected behavior**
Media Queries should be respected for the `anchor-name`

**Browsers tested on**
Safari

jgerigmeyer:
Fixed in #222

Issue #216, opened 2024-07-18
Document and re-organize codebase


jgerigmeyer:
There currently hasn't been a whole lot of thought put into which functions live in `parse.js` vs `utils.js`, and many of the files are quite large. It would be valuable to implement a better file organization structure, as well as add documentation to every function.

PR #222, opened 2024-07-24
Respect CSS cascade when determining `anchor-name`

### LINKED ISSUE #218: [BUG] `anchor-name` doesn't respect CSS cascade
**Describe the bug**
The `anchor-name` property doesn't respect the CSS cascade. For example:

```html
.anchor {
  anchor-name: --name;
}
.anchor {
  anchor-name: --other-name;
}
.positioned {
  position: absolute;
  top: anchor(--name, bottom); // this should not anchor to `.anchor`, but currently does.
}
```

**To Reproduce**
I created a test demonstrating the issue here: https://github.com/oddbird/css-anchor-positioning/compare/main...mmalerba:css-anchor-positioning:cascade

**Expected behavior**
The `anchor` function should only match the `anchor-name` that is actually applied by the cascade

**Additional context**
This can be solved by shifting `anchor-name` into a custom property, but the custom property inherits by default, so we would either have to disable that with `@property` (potentially unacceptable due to browser support?) or just be careful to verify we're not reading inherited values. This is how I handled a similar issue for `anchor-scope`: [`32f04d8` (#217)](https://github.com/oddbird/css-anchor-positioning/pull/217/commits/32f04d88e745829ba0b2fcd42f98612ad5c9fdea)

mmalerba:
Fixes #218, #212

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -1,0 +1,165 @@
+import { type VirtualElement } from '@floating-ui/dom';
+import { nanoid } from 'nanoid/non-secure';
+import { SHIFTED_PROPERTIES } from './utils.js';
+
+export interface Selector {
+  selector: string;
+  elementPart: string;
+  pseudoElementPart?: string;
+}
+
+export interface PseudoElement extends VirtualElement {
+  fakePseudoElement: HTMLElement;
+  computedStyle: CSSStyleDeclaration;
+  removeFakePseudoElement(): void;
+}
+
+export function getCSSPropertyValue(
+  el: HTMLElement | PseudoElement,
+  prop: string,
+) {
+  prop = SHIFTED_PROPERTIES[prop] ?? prop;
+  const computedStyle =
+    el instanceof HTMLElement ? getComputedStyle(el) : el.computedStyle;
+  return computedStyle.getPropertyValue(prop).trim();
+}
+
+// Given an element and CSS style property,
+// checks if the CSS property equals a certain value
+export function hasStyle(
+  element: HTMLElement | PseudoElement,
+  cssProperty: string,
+  value: string,
+) {
+  return getCSSPropertyValue(element, cssProperty) === value;
+}
+
+function createFakePseudoElement(
+  element: HTMLElement,
+  { selector, pseudoElementPart }: Selector,
+) {
+  // Floating UI needs `Element.getBoundingClientRect` to calculate the position for the anchored element,
+  // since there isn't a way to get it for pseudo-elements;
+  // we create a temporary "fake pseudo-element" that we use as reference.
+  const computedStyle = getComputedStyle(element, pseudoElementPart);
+  const fakePseudoElement = document.createElement('div');
+  const sheet = document.createElement('style');
+
+  fakePseudoElement.id = `fake-pseudo-element-${nanoid()}`;
+
+  // Copy styles from pseudo-element to the "fake pseudo-element", `.cssText` does not work on Firefox.
+  for (const property of Array.from(computedStyle)) {
+    const value = computedStyle.getPropertyValue(property);
+    fakePseudoElement.style.setProperty(property, value);
+  }
+
+  // For the `content` property, since normal elements don't have it,
+  // we add the content to a pseudo-element of the "fake pseudo-element".
+  sheet.textContent += `#${fakePseudoElement.id}${pseudoElementPart} { content: ${computedStyle.content}; }`;
+  // Hide the pseudo-element while the "fake pseudo-element" is visible.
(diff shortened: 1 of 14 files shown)


src/dom.ts
```diff
@@ -0,0 +1,238 @@
+import { type VirtualElement } from '@floating-ui/dom';
+import { nanoid } from 'nanoid/non-secure';
+import { SHIFTED_PROPERTIES } from './cascade.js';
```
jgerigmeyer, line 3:
```suggestion
import { nanoid } from 'nanoid/non-secure';

import { SHIFTED_PROPERTIES } from './cascade.js';
```

tests/helpers.ts
```diff
@@ -1,13 +1,26 @@
 import fs from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
+import { cascadeCSS } from '../src/cascade.js';
+import { StyleData } from '../src/utils.js';
```
jgerigmeyer, line 5:
```suggestion

import { cascadeCSS } from '../src/cascade.js';
import { type StyleData } from '../src/utils.js';
```

src/dom.ts
```diff
@@ -0,0 +1,238 @@
...
+ */
+export const enum AnchorScopeValue {
+  All = 'all',
+  None = 'none',
+}
+
+/**
+ * Gets the computed value of a CSS property for an element or pseudo-element.
+ *
+ * Note: values for properties that are not natively supported are *awlways* subject to CSS
```
jgerigmeyer, line 34:
```suggestion
 * Note: values for properties that are not natively supported are *always* subject to CSS
```

src/dom.ts
```diff
@@ -0,0 +1,238 @@
...
+ * Used instead of an HTMLElement as a handle for pseudo-elements.
+ */
+export interface PseudoElement extends VirtualElement {
+  fakePseudoElement: HTMLElement;
+  computedStyle: CSSStyleDeclaration;
+  removeFakePseudoElement(): void;
+}
+
+/**
+ * Possible values for `anchor-scope` (in addition any valid dashed identifier)
```
jgerigmeyer, line 24:
```suggestion
 * Possible values for `anchor-scope` (in addition to any valid dashed identifier)
```

src/dom.ts
```diff
@@ -0,0 +1,238 @@
...
+  prop = SHIFTED_PROPERTIES[prop] ?? prop;
+  const computedStyle =
+    el instanceof HTMLElement ? getComputedStyle(el) : el.computedStyle;
+  return computedStyle.getPropertyValue(prop).trim();
+}
+
+/**
+ * Checks whether a given element or pseudo-element has the given property value.
+ *
+ * Note: values for properties that are not natively supported are *awlways* subject to CSS
```
jgerigmeyer, line 50:
```suggestion
 * Note: values for properties that are not natively supported are *always* subject to CSS
```

src/dom.ts
```diff
@@ -0,0 +1,238 @@
...
+  document.head.append(sheet);
+
+  const insertionPoint =
+    pseudoElementPart === '::before' ? 'afterbegin' : 'beforeend';
+  element.insertAdjacentElement(insertionPoint, fakePseudoElement);
+  return { fakePseudoElement, sheet, computedStyle };
+}
+
+/**
+ * Finds the first scollable parent of the given element
```
jgerigmeyer, line 98:
```suggestion
 * Finds the first scrollable parent of the given element
```

src/dom.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    currentElement = currentElement.parentElement;
+  }
+
+  return currentElement;
+}
+
+/**
+ * Gets the scroll position of the first scrollable parent
+ * (or the scoll position of the element itself, if it is scrollable).
```
jgerigmeyer, line 117:
```suggestion
 * (or the scroll position of the element itself, if it is scrollable).
```

src/dom.ts
```diff
@@ -0,0 +1,238 @@
...
+
+      removeFakePseudoElement() {
+        fakePseudoElement.remove();
+        sheet.remove();
+      },
+
+      // For https://floating-ui.com/docs/autoupdate#ancestorscroll to work on `VirtualElement`s.
+      contextElement: element,
+
+      // https://floating-ui.com/docs/virtual-elements.
```
jgerigmeyer, line 177:
```suggestion
      // https://floating-ui.com/docs/virtual-elements
```

jgerigmeyer (changes requested):
This is great, and goes a long way toward #216 as well. A few typos and minor comments, but nothing concerning functionality. 👏

jgerigmeyer (approved):

PR #235, opened 2024-08-30
Fix @position-try with multiple selectors

### LINKED ISSUE #233: [BUG] @position-try can't be used by multiple selectors
**Describe the bug**
As introduced in #202, multiple selectors for a single position-try-fallbacks rule that uses an at position try rule is not supported.

```css
#targetA, #targetB{
  position-try-fallbacks: --my-at-rule;
}
```

Only 1 fallback will be made, but the at rule will apply to  `#targetA` and `#targetB` differently.

jgerigmeyer:
Fixes #233

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,8 +90,6 @@ following features:
   - a `position-area` as a `try-tactic`
   - Fallback does does not support anchor functions that are nested or passed
     through custom properties.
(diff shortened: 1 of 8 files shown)


jamesnw (approved):
Excellent!

PR #237, opened 2024-08-30
Autoupdate: Pass in empty reference element instead of target

### LINKED ISSUE #175: [bug] constantly updating example.
Hi, really nice polyfill.

I see this example constantly updating when it overlaps the border of the scrollable content. Wouldn't that make it slow. Is that a bug?

324408681-6ee6dd17-2fee-4e40-ba5a-46b39a43d137

### LINKED ISSUE #236: [BUG] Performance issues with `position-try-fallbacks`
**Describe the bug**

When `position-try-fallbacks` are used, logging seems to indicate that positions are calculated continuously and indefinitely, even when nothing on the screen is changing.

**To Reproduce**

In `applyPositionFallbacks` `autoUpdate`, add a `console.log('calculating fallbacks');` inside the handler fn.

**Expected behavior**

Positions should only be recalculated on scroll/resize or DOM changes.

**Additional context**

Related to #175 and #227.

jgerigmeyer:
Fixes #236
I think fixes #175 also?

### DIFF
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -4,6 +4,7 @@ import {
   type MiddlewareState,
   platform,
   type Rect,
+  type VirtualElement,
 } from '@floating-ui/dom';
 
 import { cascadeCSS } from './cascade.js';
@@ -369,7 +370,7 @@ async function applyPositionFallbacks(
     const offsetParent = await getOffsetParent(target);
 
     autoUpdate(
-      target,
+      {} as VirtualElement,
       target,
       async () => {
         // If this auto-update was triggered while the polyfill is already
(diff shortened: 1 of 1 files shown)


src/polyfill.ts
```diff
@@ -395,25 +396,25 @@ async function applyPositionFallbacks(
...
           // last successful fallback.
           if (index === fallbacks.length - 1) {
-            checking = false;
             const lastSuccessful = target.getAttribute(
               'data-anchor-polyfill-last-successful',
             );
             if (lastSuccessful) {
               target.setAttribute('data-anchor-polyfill', lastSuccessful);
             }
+            checking = false;
```
jgerigmeyer, line 412: In practice, this `checking` rigamarole seems to be entirely pointless, and never fires an update with `checking === true` 🤷

jamesnw: You're right, I also couldn't get it to fire while checking = true. However, I didn't see any clear guarantee that it wouldn't be true, and couldn't find a way to throttle CPU in a polyfillable browser. I'd say it doesn't hurt to keep it?

src/polyfill.ts
```diff
@@ -369,7 +370,7 @@ async function applyPositionFallbacks(
     const offsetParent = await getOffsetParent(target);
 
     autoUpdate(
```
jamesnw, line 372: Should we find a way to call the cleanup function this returns? Or perhaps that's part of #91 ?

jamesnw (approved):
That seems to work well! Nice change!

src/polyfill.ts
jgerigmeyer: Yes, and yes I think?

src/polyfill.ts
```diff
@@ -369,7 +370,7 @@ async function applyPositionFallbacks(
...
-      target,
+      {} as VirtualElement,
```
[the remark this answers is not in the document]
jgerigmeyer: I don't think we want to react to changes to the anchor element in this case, because we only case about whether the target is overflowing or not. This is a bit of a hacky workaround that depends on the internals of floating-ui, but I added a [comment](https://github.com/oddbird/css-anchor-positioning/commit/a66ce801c1c697290e9f50f303128673b9ff4345) that matches my current understanding.

src/polyfill.ts
jgerigmeyer: Filed an issue: https://github.com/oddbird/css-anchor-positioning/issues/240

PR #243, opened 2024-09-12
Fix #242: keep id,media,title attributes on replaced link elements

### LINKED ISSUE #242: `<link>` elements are replaced without all of their attributes
https://github.com/oddbird/css-anchor-positioning/blob/f1d3bcc126f62f302e4dbc519224bef3c9dcd035/src/transform.ts#L19
After this line, only the `rel` and `href` attributes are set. I'm sure there be dragons with using `el.cloneNode()` to copy all attribute values outright (event handlers, `integrity`, etc.), but these attributes should definitely be kept if present on the original element:
- `disabled`
- `id`
- `media`
- `title`

tuespetre:
I could not get the test to work with the `disabled` attribute (`jsdom` does not implement that attribute) so I just left it out.

### DIFF
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -12,13 +12,16 @@ export async function transformCSS(
       if (el.tagName.toLowerCase() === 'style') {
         // Handle inline stylesheets
         el.innerHTML = css;
-      } else if (el.tagName.toLowerCase() === 'link') {
+      } else if (el instanceof HTMLLinkElement) {
         // Create new link
         const blob = new Blob([css], { type: 'text/css' });
         const url = URL.createObjectURL(blob);
         const link = document.createElement('link');
         link.rel = 'stylesheet';
         link.href = url;
+        link.id = el.id;
+        link.media = el.media;
+        link.title = el.title;
         const promise = new Promise((res) => {
           link.onload = res;
         });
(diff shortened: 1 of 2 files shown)


src/transform.ts
```diff
@@ -12,13 +12,16 @@ export async function transformCSS(
...
         const blob = new Blob([css], { type: 'text/css' });
         const url = URL.createObjectURL(blob);
         const link = document.createElement('link');
         link.rel = 'stylesheet';
         link.href = url;
+        link.id = el.id;
+        link.media = el.media;
+        link.title = el.title;
         const promise = new Promise((res) => {
           link.onload = res;
```
jgerigmeyer, line 26: I'm fine including `disabled` here without a test, but we'll need to adjust the promise since `onload` won't fire for a disabled link. Something e.g.:

```suggestion
        link.disabled = el.disabled;
        const promise = new Promise((res) => {
          if (link.disabled) {
            res(true);
          } else {
            link.onload = res;
          }
```

jgerigmeyer (approved):

jamesnw (approved):
I was going to recommend using `el.getAttributes()` to get a fuller list, but as you point out in the issue, there's going to need to be some more intentional thought put in on which ones we actually don't want to copy. I'm happy if we look into that on a subsequent PR.

src/transform.ts
tuespetre: I was thinking about this earlier too. I wonder if it might be better to change `fetchCSS` to just ignore stylesheets with `disabled` (both `link` and `style`)

jgerigmeyer: Let's merge this, and could you open a new issue to consider what to do with `disabled` link/style elements?

PR #244, opened 2024-09-13
Fallbacks: flip anchor functions that are nested in other function calls

tuespetre:
Not aware of an existing open issue for this, but I encountered it today.

### DIFF
--- a/src/fallback.ts
+++ b/src/fallback.ts
@@ -24,6 +24,7 @@ import {
   getAST,
   getSelectors,
   INSTANCE_UUID,
+  isAnchorFunction,
   splitCommaList,
   type StyleData,
 } from './utils.js';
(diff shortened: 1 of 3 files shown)


src/fallback.ts
```diff
@@ -421,15 +420,19 @@ export function applyTryTacticToBlock(
       declarations[key] ??= 'revert';
     }
 
-    // todo: This does not support anchor functions that are nested or passed
-    // through custom properties.
-    if (isAnchorFunction(valueAst.children.first)) {
-      valueAst.children.first.children.forEach((item) => {
-        if (isIdentifier(item) && isAnchorSide(item.name)) {
-          item.name = mapAnchorSide(item.name, tactic);
+    // todo: This does not support anchor functions that are passed through custom properties.
```
jgerigmeyer, line 423:
```suggestion
    // todo: This does not support percentage anchor-side values, nor anchor
    // functions that are passed through custom properties.
```

jgerigmeyer (approved):
Thanks!

src/fallback.ts
```diff
@@ -421,15 +420,19 @@ export function applyTryTacticToBlock(
...
+    csstree.walk(valueAst, {
+      visit: 'Function',
+      enter(node) {
+        if (node.name === 'anchor') {
```
jamesnw, line 427:
```suggestion
        if (isAnchorFunction(node)) {
```
While `isAnchorFunction` will recheck that it's a function despite the fact that we already know it's a function based on `visit: 'Function'`, I'd prefer to keep using the shared util.

jamesnw (approved):
Can you also update the Limitations section in the README.md to reflect this improvement? 

Thanks for doing this!

Issue #228, opened 2024-08-28
The polyfill won’t work with `CSSStyleSheet()` or `CSSStyleDeclaration()`


marchbox:
**Is your feature request related to how popovers work in general?**

No.

**Is your feature request related to how this polyfill works? Please describe.**

When using `*.adoptedStyleSheets` or `*.style.setProperty()` to add CSS Anchor Positioning declarations, in unsupported browsers, these declarations will be ignored/removed since they are not recognized CSS properties and values. As a result, this polyfill won't work because those declarations aren't there in the first place.

The current work around is to use `*.setAttribute('style', '...')` or creating a `style` element with the Anchor Positioning declarations as its text content. But this work around could trigger content security policy issues.

**Describe the solution you'd like**

Unclear.

**Additional context**

We are currently building some custom elements (using Popover API) that would benefit from CSS Anchor Positioning, so we are investigating if using the polyfill would help in unsupported browsers. We understand the cross-root referencing issue, so in the custom element, we creates a new `CSSStyleSheet()` object and inject that into `document.adoptedStyleSheets`, e.g.

```css
#trigger-123 {
  anchor-name: --trigger-123;
}
#popover-123 {
  left: anchor(--trigger-123, left);
  top: anchor(--trigger-123, bottom);
}
```

But in unsupported browsers, this style sheet would just be:

```css
#trigger-123 {}
#popover-123 {}
```

We are wondering if there is a solution for this use case, or rather this is out of the scope of this polyfill.

jgerigmeyer:
Hm, good question. It's definitely related to this polyfill, but somewhat tangential to the work of the polyfill itself -- more generally "How can I define styles for custom elements in browsers that don't understand them as styles?".

I'm not thinking of anything other than `<link>` or the workarounds you outlined (`createElement('style')` with `textContent`). @jamesnw @mirisuzanne @mmalerba Any ideas?

mmalerba:
I think if we really wanted to be fancy we could try to monkey-patch the relevant APIs to intercept these styles, for example:
```ts
const realReplaceSync = CSSStyleSheet.prototype.replaceSync;
CSSStyleSheet.prototype.replaceSync = function(css) => {
    addToPolyfill(css);
    return realReplaceSync.apply(this, arguments);
}
```

[...]

Issue #255, opened 2024-10-03
[BUG] Polyfill fails if a `<link>` fails to load or returns non-CSS content


rkh:
Great work on the polyfill! Haven't gotten it working yet, but hopefully I'll manage to.

**Describe the bug**
It is impossible for me to apply the polyfill. Getting the same error (different stack trace depending on the method used) when applying via Vite in dev mode, Vite bundle, or loading from unpkg (not my preference, as I don't want to add that to the CSP for production).

Example stacktrace:

```
Uncaught (in promise) TypeError: e.children is undefined
    ir utils.ts:70
    Tm parse.ts:306
    h create.js:169
    Xr create.js:108
    h create.js:180
    k create.js:195
    reduce List.js:174
    Xr create.js:103
    h create.js:180
    l create.js:239
    Tm parse.ts:304
    B @oddbird_css-anchor-positioning_fn.js:34
    B @oddbird_css-anchor-positioning_fn.js:20
    Tm parse.ts:294
    qm polyfill.ts:460
    o @oddbird_css-anchor-positioning_fn.js:23
    promise callback*c @oddbird_css-anchor-positioning_fn.js:33
    o @oddbird_css-anchor-positioning_fn.js:23
    promise callback*c @oddbird_css-anchor-positioning_fn.js:33
    o @oddbird_css-anchor-positioning_fn.js:23
    promise callback*c @oddbird_css-anchor-positioning_fn.js:33
    B @oddbird_css-anchor-positioning_fn.js:34
    B @oddbird_css-anchor-positioning_fn.js:20
    qm polyfill.ts:446
    setTimeout handler* application.ts:42
    onReady shim.ts:13
    async* application.ts:34
```

This happens no matter if I load the Polyfill before or after initializing the rest of the app, no matter if any CSS rules actually specify any anchor properties.

The line numbers seem to be a little off (possibly a sourcemap issue?). I assume the line in question is [line 72](https://github.com/oddbird/css-anchor-positioning/blob/6f2bff4519bb9f5ccaf07d16b20e34220706e981/src/utils.ts#L72).

FYI the [example page](https://anchor-polyfill.netlify.app/) is working for me.

**To Reproduce**
It's a fairly complex app; I'll look into creating an example repo in the coming days, but I wanted to see if, maybe, from the stack trace, this is already something you can spot.

Code I'm using:

```ts
import polyfill from '@oddbird/css-anchor-positioning/fn'
[...]

Issue #253, opened 2024-10-02
[BUG] Stylesheet `<link>` URLs are swapped for `blob`s in v0.2.0


tysongach:
**Describe the bug**

I have an existing Rails application that I just upgraded from v0.1.0 to v0.2.0 and I’m now getting security warnings in the console:

```
Security Error: Content at http://localhost:3000/ may not load data from blob:http://localhost:3000/c359010b-028d-4dc6-823e-0ec6a8f4c19a.
```

I’m noticing that the stylesheets linked in the `<head>` have `blob:` URLs - I’m not sure if this is new:

```html
<link rel="stylesheet" href="blob:http://localhost:3000/c359010b-028d-4dc6-823e-0ec6a8f4c19a">
```

This issue is also causing my fonts not to load.

If I remove `import "@oddbird/css-anchor-positioning"` the issue goes away.

**To Reproduce**
Use the polyfill. I’ve not determined if this is isolated to Rails or [Turbo](https://turbo.hotwired.dev/).

**Expected behavior**
No console security error.

**Browsers tested on**
macOS, Firefox 131

jgerigmeyer:
When styles are polyfilled, the URL is replaced with a `blob:` URL -- this is intentional, and not a change from v0.1.*. I'm not sure why you were seeing some links polyfilled and others not (unless their contents no longer included CSS that required polyfilling?), but this specific error should be fixed in [v0.3.0](https://github.com/oddbird/css-anchor-positioning/releases/tag/v0.3.0).

Issue #245, opened 2024-09-13
Request: opt-in mechanism for stylesheets


tuespetre:
I've got stylesheets that I do not want the polyfill to process. For instance, my project's main stylesheet, and an auxiliary stylesheet with a `media` attribute for conditional loading. These each have a couple of odd and end rules in it that have properties like `inset: auto;` or `top: 0;`, so the polyfill is processing them and generating custom properties for them, even though I know for a fact that these rules will never be relevant to anything I am using anchor positioning with. I want to avoid the re-rendering caused by this, but I would rather not sequester those odd couple of rules into separate stylesheets, they are rather happy just where they are 😄 

What I would like to see is a configuration option that puts the polyfill into "opt-in mode", by allowing me to specify an attribute name so that only stylesheets with that attribute name will be processed.

```
// Still accept the single boolean argument for compatibility:
polyfill(true);

// Also accept an object argument...
polyfill({ 
    updateAnchorOnAnimationFrame: true
});

// ... and add the new option to the object argument for "opt-in mode".
polyfill({
    optInAttribute: 'data-polyfill-anchor'
}); 

// Also allow a global variable to specify the object argument.
window.ANCHOR_POSITIONING_POLYFILL_OPTIONS = {
    updateAnchorOnAnimationFrame: true,
    optInAttribute: 'data-polyfill-anchor'
};
```

Let me know what y'all think

jgerigmeyer:
This is released in [v0.3.0](https://github.com/oddbird/css-anchor-positioning/releases/tag/v0.3.0).

PR #256, opened 2024-10-08
Allow polyfilling imperatively for a single style element

marchbox:
## Description

This change provides an alternative way for authors to apply polyfill: instead of letting the polyfill to lookup stylesheets and 
apply the polyfill, the second argument takes a `<style>` element and the function can apply polyfill based on its content.

This is particularly useful for components like custom elements, as often the stylesheets are added dynamically to the document, and the currently "auto" polyfilling makes it challenging to apply the changes at the right time.

## Related Issue(s)

https://github.com/oddbird/css-anchor-positioning/issues/228

### DIFF
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -26,7 +26,6 @@ import {
   type SizingProperty,
 } from './syntax.js';
 import { transformCSS } from './transform.js';
-import type { StyleData } from './utils.js';
 
 const platformWithCache = { ...platform, _c: new Map() };
 
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -26,6 +26,7 @@ import {
   type SizingProperty,
 } from './syntax.js';
 import { transformCSS } from './transform.js';
+import type { StyleData } from './utils.js';
 
 const platformWithCache = { ...platform, _c: new Map() };
 
@@ -443,13 +444,14 @@ async function position(rules: AnchorPositions, useAnimationFrame = false) {
   }
 }
 
-export async function polyfill(animationFrame?: boolean) {
+export async function polyfill(animationFrame?: boolean, el?: HTMLStyleElement) {
   const useAnimationFrame =
     animationFrame === undefined
       ? Boolean(window.UPDATE_ANCHOR_ON_ANIMATION_FRAME)
       : animationFrame;
+
   // fetch CSS from stylesheet and inline style
-  let styleData = await fetchCSS();
+  let styleData = el ? [{el, css: el.textContent ?? ''}] : await fetchCSS();
 
   // pre parse CSS styles that we need to cascade
   const cascadeCausedChanges = await cascadeCSS(styleData);


src/polyfill.ts
```diff
@@ -443,13 +443,14 @@ async function position(rules: AnchorPositions, useAnimationFrame = false) {
...
-export async function polyfill(animationFrame?: boolean) {
+export async function polyfill(animationFrame?: boolean, el?: HTMLStyleElement) {
   const useAnimationFrame =
     animationFrame === undefined
       ? Boolean(window.UPDATE_ANCHOR_ON_ANIMATION_FRAME)
       : animationFrame;
+
   // fetch CSS from stylesheet and inline style
-  let styleData = await fetchCSS();
+  let styleData = el ? [{el, css: el.textContent ?? ''}] : await fetchCSS();
```
jgerigmeyer, line 453: Elsewhere we're using `innerHTML`, but I'm not sure if there's a great argument for one over the other.

src/polyfill.ts
```diff
@@ -443,13 +443,14 @@ async function position(rules: AnchorPositions, useAnimationFrame = false) {
   }
 }
 
-export async function polyfill(animationFrame?: boolean) {
+export async function polyfill(animationFrame?: boolean, el?: HTMLStyleElement) {
```
jgerigmeyer, line 446: We should have done this from the beginning, but I think we should switch this from 2 argument to 1 options object. For backwards compatibility we can check if `typeof options === 'boolean'`, but otherwise going forward that will be both clearer and more flexible.

jgerigmeyer (commented):
I'm okay with this approach. I wonder if we want to accept an array of elements, and we could also support `<link>` tags as well as `<style>`? Basically using the same functionality as the existing polyfill, but allowing users to opt into a specified list of CSS sources.

@jamesnw @mmalerba What do you think?

jamesnw:
> I'm okay with this approach. I wonder if we want to accept an array of elements, and we could also support `<link>` tags as well as `<style>`? Basically using the same functionality as the existing polyfill, but allowing users to opt into a specified list of CSS sources.

I think an array that supports `<link>` tags as well as `<style>` makes a lot of sense. 

This does omit inline styles- could we add another boolean flag `includeInline` that defaults to true, and then adds the response of `fetch:fetchInlineStyles()`? Or since the inline style impact is scoped to just ones that include `anchor`, should we just always add the inline styles?

It would be nice to have a demo, but that's a bit of a challenge in the existing single page setup.

marchbox:
Thanks for the comments. I'll add them in.

> I think an array that supports <link> tags as well as <style> makes a lot of sense.
>
> This does omit inline styles- could we add another boolean flag includeInline that defaults to true, and then adds the response of fetch:fetchInlineStyles()? Or since the inline style impact is scoped to just ones that include anchor, should we just always add the inline styles?

Sounds good. I'll pass in the element array into both `fetchCSS()` and `fetchInlineStyle()` to build the `styleData` object.

marchbox:
I updated the code and it's ready for review:

* Changed the argument of `polyfill()` to accept an object for configuration, with backward compatibility
* You can pass in an array of elements for polyfilling, and imperative polyfill uses the same functions as auto polyfill
* Added demo and tests

Also, maybe I should rename "imperative" to "manual"? Let me know if "manual" sounds better and less confusing.

jamesnw (changes requested):
Excellent work, this is quite helpful! A couple quick notes-

- It looks like there are a few linting issues, you can run `npm run lint` to fix or surface.
- I do prefer `manual` over `imperative` slightly, but it's not a blocker.
- Is it a legitimate use case for people to pass in explicit `<style>` or `<link>` elements but want to polyfill all inline styles automatically? If so, that currently isn't possible. 
- If you click the imperative button, and then apply the polyfill to the whole page, the polyfill crashes. This is somewhat related to #255 as well as #91, so I'm not concerned here. 
- Do you have any concerns about running this on 2 discrete sets of elements separately? It seems like it should work, as long as there aren't references between the 2 sets of elements.

Overall, thanks for the contribution!

src/fetch.ts
```diff
@@ -52,12 +61,16 @@ function fetchInlineStyles() {
...
+  const targetElements: HTMLElement[] = Array.isArray(elements)
+    ? elements
+    : Array.from(document.querySelectorAll('link, style'));
   const sources: Partial<StyleData>[] = [];
 
-  elements.forEach((el) => {
+  targetElements.forEach((el) => {
+    if (!(el instanceof HTMLElement)) {
+      return;
+    }
```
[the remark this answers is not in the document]
marchbox: Ah definitely like the filter better! Thanks!

index.html
```diff
@@ -78,6 +84,77 @@
...
+              document.getElementById('my-target-imperative-inline-style'),
+            ],
+          }).then((rules) => {
+            imperativeBtn.innerText = 'Polyfill Applied';
+            imperativeBtn.setAttribute('disabled', '');
+            console.log(rules);
+          });
+        });
+      } else {
+        prepareImperativePolyfill();
```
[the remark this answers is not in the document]
marchbox: Yeah sounds good! I'll update this.

marchbox:
> Is it a legitimate use case for people to pass in explicit <style> or <link> elements but want to polyfill all inline styles automatically? If so, that currently isn't possible.

I think that's a valid use case. I'll add this. My only concern is, if the `elements` already contains some elements that have inline styles, then you pass in `includeInlineStyles: false`, an author may expect the elements with inline styles in `elements` not get polyfilled. But this can probably be clarified by documentation. Or maybe we can give `includeInlineStyles` a different name for more clarity.

> Do you have any concerns about running this on 2 discrete sets of elements separately? It seems like it should work, as long as there aren't references between the 2 sets of elements.

Yeah this should work, and this is our intended use since multiple components in our library would need to manual polyfill separately. I also added some test coverage: https://github.com/oddbird/css-anchor-positioning/pull/256/commits/912c553c6e03502324ce9e4575e766321a4cff06.

The only caveat though, is that, if the elements in different sets rely on the same `anchor-name`, then all `polyfill()` call need to include the element that contains this `anchor-name` declaration. I'm not sure if there's a map somewhere that keeps a record of the anchor name and the generated custom property name. It would be nice to be able to reference an already transformed anchor name.

marchbox:
Added `includeInlineStyles`: https://github.com/oddbird/css-anchor-positioning/pull/256/commits/798561f9dbe8d38ac5c740d3fb3aaafd2de0b0b3

jamesnw:
> > Is it a legitimate use case for people to pass in explicit <style> or  elements but want to polyfill all inline styles automatically? If so, that currently isn't possible.
> 
> I think that's a valid use case. I'll add this. My only concern is, if the `elements` already contains some elements that have inline styles, then you pass in `includeInlineStyles: false`, an author may expect the elements with inline styles in `elements` not get polyfilled. But this can probably be clarified by documentation. Or maybe we can give `includeInlineStyles` a different name for more clarity.

I have a slightly different thought on how this could work that may make it more clear (either way, documentation will be useful). 

We could call it `discoverInlineStyles` or `transformInlineStyles` or ..., have it default to true, and separate it from whether or not `elements` are passed in. In other words, we default to always finding all the inline styles, but a user can opt out because they are passing in the elements they know have inline styles OR because they know they don't have any inline styles with `*anchor`.

Would that be simpler for users?

index.html
```diff
@@ -1001,6 +1119,87 @@ <h2>
...
+  }
+&lt;/style&gt;
+&lt;style id="my-style-manual-style-el"&gt;
+  #my-target-manual-style-el {
+    position: absolute;
+    bottom: anchor(--my-anchor-manual top);
+    right: anchor(--my-anchor-manual left);
+  }
+&lt;/style&gt;
+&lt;link rel="stylesheet" href="/anchor-manual.css" id="my-style-manual-link-el"&gt;
```
jgerigmeyer, line 1168:
```suggestion
&lt;link rel="stylesheet" href="/anchor-manual.css" id="my-style-manual-link-el" /&gt;
```

index.html
```diff
@@ -1001,6 +1119,87 @@ <h2>
...
+&lt;div id="my-target-manual-link-el" class="target"&gt;...&lt;/div&gt;
+&lt;div id="my-target-manual-inline-style" class="target"
+  style="position: absolute;
+    top: anchor(--my-anchor-manual bottom);
+    left: anchor(--my-anchor-manual right);"
+&gt;...&lt;/div&gt;
+
+&lt;script&gt;
+  polyfill({
+    useAnimationFrame: true,
```
jgerigmeyer, line 1190: Same question as above...

```suggestion
```

index.html
```diff
@@ -78,6 +85,117 @@
...
+      const manualBtn = document.getElementById('apply-polyfill-manually');
+      if (SUPPORTS_ANCHOR_POSITIONING) {
+        manualBtn.innerText = 'Load Anchor Positioning CSS';
+      }
+      manualBtn.addEventListener('click', () => {
+        prepareManualPolyfill();
+
+        if (!SUPPORTS_ANCHOR_POSITIONING) {
+          polyfill({
+            useAnimationFrame: true,
```
jgerigmeyer, line 180: Is there a reason to set `useAnimationFrame` here? I'd generally recommend people not use it unless they need to.

```suggestion
```

index.html
```diff
@@ -1001,6 +1119,87 @@ <h2>
   top: anchor(--my-anchor-media-query top);
   right: anchor(--my-anchor-media-query right);
 }</code></pre>
+    </section>
+    <section id="anchor-manual" class="demo-item" style="position: relative">
+      <h2>
+        <a href="#manual" aria-hidden="true">🔗</a>
+        Manually apply polyfill
```
jgerigmeyer, line 1126: I wonder if there's a clearer description for the intent of this change. Maybe:

```suggestion
        Polyfill applied to only specified styles
```

Or something along those lines?

src/fetch.ts
```diff
@@ -31,45 +31,64 @@ async function fetchLinkedStylesheets(
...
 // random UUID value, and then formats the styles in the same manner as CSS from
 // style tags.
-function fetchInlineStyles() {
-  const elementsWithInlineAnchorStyles: NodeListOf<HTMLElement> =
-    document.querySelectorAll('[style*="anchor"]');
+function fetchInlineStyles(elements?: HTMLElement[]) {
+  const elementsWithInlineAnchorStyles: HTMLElement[] = elements
+    ? elements.filter((el) =>
+        el?.matches?.(ELEMENTS_WITH_INLINE_ANCHOR_STYLES_QUERY),
+      )
```
jgerigmeyer, line 43: It doesn't hurt anything, but I'm curious why the optional chaining is needed here.

```suggestion
    ? elements.filter((el) =>
        el.matches(ELEMENTS_WITH_INLINE_ANCHOR_STYLES_QUERY),
      )
```

src/fetch.ts
```diff
@@ -31,45 +31,64 @@ async function fetchLinkedStylesheets(
...
-  elementsWithInlineAnchorStyles.forEach((el) => {
-    const selector = nanoid(12);
-    const dataAttribute = 'data-has-inline-styles';
-    el.setAttribute(dataAttribute, selector);
-    const styles = el.getAttribute('style');
-    const css = `[${dataAttribute}="${selector}"] { ${styles} }`;
-    inlineStyles.push({ el, css });
-  });
+  elementsWithInlineAnchorStyles
+    .filter((el) => el instanceof HTMLElement)
```
jgerigmeyer, line 50: This is also more defensive than strictly necessary, right? I'm fine leaving it, just want to be sure I'm understanding the intent.

src/polyfill.ts
```diff
@@ -443,13 +443,50 @@ async function position(rules: AnchorPositions, useAnimationFrame = false) {
   }
 }
 
-export async function polyfill(animationFrame?: boolean) {
+export interface PolyfillOptions {
+  // Whether to use `requestAnimationFrame()` when updating target elements’
+  // positions
+  useAnimationFrame?: boolean;
+
+  // An array of explictily targeted elements to polyfill
```
jgerigmeyer, line 451:
```suggestion
  // An array of explicitly targeted elements to polyfill
```

src/polyfill.ts
```diff
@@ -443,13 +443,50 @@ async function position(rules: AnchorPositions, useAnimationFrame = false) {
...
+  const options =
+    typeof useAnimationFrameOrOption === 'boolean'
+      ? { useAnimationFrame: useAnimationFrameOrOption }
+      : useAnimationFrameOrOption;
   const useAnimationFrame =
-    animationFrame === undefined
+    options.useAnimationFrame === undefined
       ? Boolean(window.UPDATE_ANCHOR_ON_ANIMATION_FRAME)
-      : animationFrame;
+      : options.useAnimationFrame;
```
jgerigmeyer, line 473: In #245 there was a suggestion to convert this `window.UPDATE_ANCHOR_ON_ANIMATION_FRAME` option into a broader `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS` object, which would allow using the new `elements` and `excludeInlineStyles` with the default build, and would match the options object pattern we're using here. It seems pretty straightforward so I'm inclined to do that -- any objections to that idea?

src/polyfill.ts
```diff
@@ -443,13 +443,50 @@ async function position(rules: AnchorPositions, useAnimationFrame = false) {
...
+
+  if (options.elements && !Array.isArray(options.elements)) {
```
jgerigmeyer, line 475:
```suggestion
  if (!Array.isArray(options.elements)) {
```

src/fetch.ts
```diff
@@ -31,45 +31,64 @@ async function fetchLinkedStylesheets(
...
-      sources.push({ el, css: el.innerHTML });
-    }
-  });
+    });
+
+  const elementsForInlines = excludeInlineStyles
+    ? elements?.length
+      ? elements // Only elements in `elements`
+      : [] // No elements
+    : undefined; // All elements
```
jgerigmeyer, line 89: Is there a benefit to checking for `elements?.length`?

```suggestion
  const elementsForInlines = excludeInlineStyles ? elements : undefined;
```

tests/unit/fetch.test.ts
```diff
@@ -97,3 +97,122 @@ describe('fetch inline styles', () => {
...
+    const styleData = await fetchCSS([
+      el1,
+      el2,
+      el3,
+      el4,
+      // should be ignore
+      el5,
+      // @ts-expect-error should be ignored
+      undefined,
+      // @ts-expect-error should be ignore
```
jgerigmeyer, line 190:
```suggestion
      // should be ignored
      el5,
      // @ts-expect-error should be ignored
      undefined,
      // @ts-expect-error should be ignored
```

jgerigmeyer (commented):
I love this set of changes -- thanks again for the contribution!

Looks like there are two failing tests. I left mostly minor nit comments, and one question/suggestion for expanding the options API.

jgerigmeyer:
> * If you click the imperative button, and then apply the polyfill to the whole page, the polyfill crashes. This is somewhat related to [[BUG] Polyfill fails if a `<link>` fails to load or returns non-CSS content #255](https://github.com/oddbird/css-anchor-positioning/issues/255) as well as [applying the polyfill to dynamic elements #91](https://github.com/oddbird/css-anchor-positioning/issues/91), so I'm not concerned here.

I'm seeing this too. I think it relates to #253, as the error I see seems to be when the polyfill tries to fetch the already-polyfilled style:

```
Security Error: Content at http://localhost:3000/ may not load data from blob:http://localhost:3000/8749e98f-f55a-4153-9249-f6bba301e298.
```

I'm not sure if there's an easy way to read/fetch a blob url, or whether we should explicitly just skip those links?

index.html
marchbox: No particular reasons, I'll remove them.

index.html
marchbox: How about "Manually apply polyfill to specific styles"?

src/fetch.ts
marchbox: I think I probably used them before I added type checking to make sure they will indeed be `HTMLElement`s. I'll remove these.

src/fetch.ts
marchbox: Since the previous check in `normalizePolyfillOptions()` in `polyfill.ts` only makes sure that `elements` is either an array or `undefined`, so I want to make sure no JS error would be thrown if an author acceidentally passed in a non `HTMLElement` array member.

src/polyfill.ts
marchbox: Nice catch!

src/polyfill.ts
marchbox: Ah nice, no objection here. I think I can incorporate that in this PR.

src/fetch.ts
marchbox: This is cover this case:

```js
polyfill({ excludeInlineStyles: true });
```

In this case, the author opted out of auto polyfilling inline styles, but didn't pass in an `elements` array. Without this check, `elements` would be `undefined`, and `fetchInlineStyles()` would polyfill all inline styles.

marchbox:
> > * If you click the imperative button, and then apply the polyfill to the whole page, the polyfill crashes. This is somewhat related to [[BUG] Polyfill fails if a `<link>` fails to load or returns non-CSS content #255](https://github.com/oddbird/css-anchor-positioning/issues/255) as well as [applying the polyfill to dynamic elements #91](https://github.com/oddbird/css-anchor-positioning/issues/91), so I'm not concerned here.
> 
> I'm seeing this too. I think it relates to #253, as the error I see seems to be when the polyfill tries to fetch the already-polyfilled style:
> 
> ```
> Security Error: Content at http://localhost:3000/ may not load data from blob:http://localhost:3000/8749e98f-f55a-4153-9249-f6bba301e298.
> ```
> 
> I'm not sure if there's an easy way to read/fetch a blob url, or whether we should explicitly just skip those links?

Not quite a fix but this would at least let other styles being polyfilled:

```ts
try {
  // fetch css and add to array
  const response = await fetch(data.url.toString());
  const css = await response.text();
  return { ...data, css } as StyleData;
} catch {
  return { ...data, css: ''} as StyleData;
}
```

marchbox:
@jamesnw  @jgerigmeyer Thanks for the reviews, both. Have addressed all comments. Ready for another round :)

src/fetch.ts
jgerigmeyer: Ah, I see! I might find it slightly easier to read as:

```suggestion
  const elementsForInlines = excludeInlineStyles ? (elements ?? []) : undefined;
```

But it's also fine as-is. 👍

jgerigmeyer (approved):
> Not quite a fix but this would at least let other styles be polyfilled:
> 
> ```ts
> try {
>   // fetch css and add to array
>   const response = await fetch(data.url.toString());
>   const css = await response.text();
>   return { ...data, css } as StyleData;
> } catch {
>   return { ...data, css: ''} as StyleData;
> }
> ```

I think this might already be fixed by #258? Regardless, I don't think we need to address that in this PR.

Thanks again for this contribution! 🎉

jamesnw (approved):
This works great! @marchbox I opened PR #259  to update the docs with the new configuration options- I'd appreciate your eyes on that to make sure it makes sense. Thanks!

marchbox:
> This works great! @marchbox I opened PR #259 to update the docs with the new configuration options- I'd appreciate your eyes on that to make sure it makes sense. Thanks!

Thank you! Will take a look at the documentation changes today.

PR #257, opened 2024-10-11
Replace the <link> without flashing the css

mariusGundersen:
This fixes the flash of unstyled content (FOUC) by inserting the new stylesheet before the old one is removed.

### DIFF
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -25,9 +25,10 @@ export async function transformCSS(
         const promise = new Promise((res) => {
           link.onload = res;
         });
-        el.replaceWith(link);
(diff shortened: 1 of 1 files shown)


jamesnw (changes requested):
Overall, I think this makes sense. Thanks for this! The momentary double styles will be less impactful less momentary no styles.

The only change I see is with the tests, which I think will be easy to fix- it looks like since both `link`s are in the DOM, `querySelector` selects the old one, so the `load` event is called on the wrong element.

jamesnw (approved):
That change fixed the tests. Thanks for this contribution!

jgerigmeyer (approved):

PR #258, opened 2024-10-11
Warn if mime type is not CSS

jamesnw:
## Description
If a `<link>` fails to load, we should recover gracefully. This may be because it's a 404 or because it's being loaded as CSS-in-JS for HMR. It possibly is due to a misconfigured server. Ideally we would confirm if it's actually CSS instead of just relying on mimetype, but there's not a good way of checking that.

## Related Issue(s)
#255

## Steps to test/reproduce

- Load the demo in a polyfilled browser, see that `fake.css` fails to load initially. 
- Apply the polyfill.
- See that the polyfill applies successfully
- A console warning shows what the error is

## Show me
https://deploy-preview-258--anchor-polyfill.netlify.app/

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,6 +16,8 @@
       defer
     ></script>
     <link rel="stylesheet" href="/demo.css" />
+    <!-- Included to test invalid  -->
+    <link rel="stylesheet" href="/fake.css" />
     <link rel="stylesheet" href="/anchor.css" />
     <link rel="stylesheet" href="/anchor-positioning.css" />
     <link rel="stylesheet" href="/anchor-popover.css" />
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -16,6 +16,8 @@
       defer
     ></script>
     <link rel="stylesheet" href="/demo.css" />
+    <!-- Included to test invalid  -->
```
jgerigmeyer, line 19:
```suggestion
    <!-- Included to test invalid stylesheets -->
```

jgerigmeyer (approved):

index.html
jamesnw: @jgerigmeyer I wasn't sure- do we actually want to keep this in? It makes the tests a bit noisier.

jgerigmeyer: I'd lean toward keeping it -- the cost of noisy tests seems low to me 🤷‍♂️

PR #260, opened 2024-10-14
Do not revoke object URL of polyfilled styles

jgerigmeyer:
## Description

After loading the polyfilled styles, we were calling [`revokeObjectURL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static) to release the URL. If the polyfill is applied again, this caused an error when trying to fetch the already-polyfilled styles from the blob URL.

## Related Issue(s)

#253 

## Steps to test/reproduce

- Open in Firefox
- Click the "Polyfill these elements" manual button
- Click the global "Apply Polyfill" button
- No error is thrown

### DIFF
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -29,7 +29,6 @@ export async function transformCSS(
         // Wait for new stylesheet to be loaded
         await promise;
         el.remove();
-        URL.revokeObjectURL(url);
         updatedObject.el = link;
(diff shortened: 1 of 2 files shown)


jamesnw (approved):
My only hesitation is whether we want to keep a list of the urls around so that we could clean them up later, but I can't think of a tangible benefit of doing so.

Issue #246, opened 2024-09-13
Handling stylesheet links that have `disabled` and/or `media`


tuespetre:
The issue:
- The browser will not load stylesheets from links with the `disabled` attribute.
- The polyfill currently will load such links, and if any changes are applied, replace them with new links without `disabled`
- Even if the polyfill is modified to preserve `disabled`, it might not be desirable to fetch the resource in the first place.
- The same might be said for links with a `media` attribute when the media query is not currently satisfied.

My rough idea to address the issue is basically this: 
- For any stylesheet link that the polyfill would process:
    - Check if it is `disabled`. If so, add a `MutationObserver` with `{ attributeFilter: ['disabled'] }` that will handle processing the stylesheet if it were to become no longer disabled.
    - Check if it has `media`, and if so, create a `MediaQueryList`. If the media query list does not match, add an event listener for its `change` event that will handle processing the stylesheet if it were to become applicable according the media query.

PR #262, opened 2024-10-14
Do not fetch disabled links, but copy over all attributes to new link.

### LINKED ISSUE #254: [BUG] Data attributes get stripped from stylesheet `<link>`s
**Describe the bug**

The polyfill appears to be stripping `data` attributes.

Without the polyfill, my `link` looks like this:

```html
<link rel="stylesheet" href="/assets/application-3282d104.css" data-turbo-track="reload" media="screen">
```

And when I add the polyfill, I get this:

```html
<link rel="stylesheet" href="blob:http://localhost:3000/ae9daae1-2ee9-44e5-914b-2a829a711088">
```

Note that `media` also gets stripped and should be kept, but I believe https://github.com/oddbird/css-anchor-positioning/pull/243 resolves that. Looking at those code changes, I don’t believe it accounts for `data` attributes, though.

**To Reproduce**
Use the polyfill.

**Expected behavior**
Existing `data` attributes should still exist on the `link` when polyfill is applied.

**Browsers tested on**
macOS, Firefox 131

**Additional context**
I’m seeing this issue in a Rails application with [Turbo](https://turbo.hotwired.dev).

jgerigmeyer:
## Description

This doesn't yet include the full fix for #246 because we don't have a robust solution for running the polyfill multiple times on the same page, but it makes two changes:

1. Links with `disabled` are now ignored entirely.
2. All attributes are copied over from the original link to the new one.

## Related Issue(s)

Fixes #254.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -25,6 +25,12 @@ async function fetchLinkedStylesheets(
       if (!data.url) {
         return data as StyleData;
       }
+      // TODO: Add MutationObserver to watch for disabled links being enabled
+      // https://github.com/oddbird/css-anchor-positioning/issues/246
+      if ((data.el as HTMLLinkElement | undefined)?.disabled) {
+        // Do not fetch or parse disabled stylesheets
+        return null;
+      }
       // fetch css and add to array
       try {
         const response = await fetch(data.url.toString());
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -17,11 +17,13 @@ export async function transformCSS(
         const blob = new Blob([css], { type: 'text/css' });
         const url = URL.createObjectURL(blob);
         const link = document.createElement('link');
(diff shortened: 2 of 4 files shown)


jamesnw (approved):
LGTM! I think in addition to the mutation observer, we may still want to separately address things like `integrity` or `referrerpolicy`, or event handlers that we don't want to bring over. These were initially raised in #242.

jgerigmeyer:
> LGTM! I think in addition to the mutation observer, we may still want to separately address things like `integrity` or `referrerpolicy`, or event handlers that we don't want to bring over. These were initially raised in #242.

Hm, good point. Do you think it would be worth compiling a list of either the attributes we want to explicitly exclude, or explicitly include? It's not perfect, but might be better than either the current implementation or this PR?

jamesnw (approved):
This looks good!

PR #271, opened 2024-11-19
Update demo site with OddBird branding

stacyk:
## Description
- Removed open props stylesheet, do we need it?
- Replaced some css variables with OddBird brand tokens
- Not sure if we will need to update the Adobe font package?

## Show me
Screenshot 2024-11-19 at 2 16 40 PM

### DIFF
--- a/public/demo.css
+++ b/public/demo.css
@@ -4,39 +4,50 @@
 }
 
 html {
-  --action: var(--action-light);
-  --action-dark: var(--pink-2);
-  --action-light: hsl(339deg 68% 37%);
-  --anchor-color: var(--yellow-6);
-  --bg: var(--gray-1);
-  --brand: var(--green-9);
+  --brand-blue: lch(38.953 23.391 229.55);
+  --brand-pink: lch(50.161 77.603 3.8969);
+  --brand-pink-light: lch(from var(--brand-pink) calc(l + 8) c h);
+  --brand-pink-dark: lch(from var(--brand-pink) calc(l - 20) c h);
+  --brand-orange: lch(70.149 72.526 55.336);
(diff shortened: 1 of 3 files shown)


jamesnw:
This one doesn't seem right (I can troubleshoot later)
image

stacyk:
> This one doesn't seem right (I can troubleshoot later) image

The padding changed since I updated the the html font-size to `calc(1em + 0.25vw)` so things adjusted. When I used padding: 0 on that example, I think it looks like what you expected it to? 
Screenshot 2024-11-19 at 4 03 11 PM

jamesnw:
@stacyk That is better, but I realized that the 50px offset wasn't doing a great job of clearly communicating what was happening, so I changed it a bit. The absolute positioning one was overlapping a bit, so I adjusted that as well.

public/demo.css
```diff
@@ -4,39 +4,52 @@
...
-  --anchor-color: var(--yellow-6);
-  --bg: var(--gray-1);
-  --brand: var(--green-9);
+  --brand-blue: lch(38.953% 23.391 229.55deg);
+  --brand-pink: lch(50.161% 77.603 3.8969deg);
+  --brand-pink-light: lch(from var(--brand-pink) calc(l + 8) c h);
+  --brand-pink-dark: lch(from var(--brand-pink) calc(l - 20) c h);
+  --brand-orange: lch(70.149% 72.526 55.336deg);
+  --brand-orange-dark: lch(from var(--brand-orange) calc(l - 20) c h);
+  --gray-1: lch(from var(--brand-blue) calc(l + 53) calc(c - 19) h);
```
mirisuzanne, line 13: 😮😮😮😮 coooool

mirisuzanne (approved):

PR #272, opened 2024-11-20
Add note about implicit support

jamesnw:
The `anchor` attribute has been removed from the standard pending conversation here- https://github.com/whatwg/html/pull/9144. Alternatively, we could remove this feature until it is supported, but I'm not sure it's worth it.

### DIFF
--- a/index.html
+++ b/index.html
@@ -435,6 +435,12 @@
       <p class="note">
         With polyfill applied: Target is positioned at the top left corner of
         the Anchor.
+        <br />
+        <br />
+        <strong>Note:</strong> The <code>anchor</code> attribute has not yet
+        been standardized, and currently does not work in non-polyfilled
+        browsers. See more conversation <a
+        href="https://github.com/whatwg/html/pull/9144">here</a>.
       </p>
       <pre><code class="language-html"
 >&lt;div id="my-implicit-anchor"&gt;Anchor&lt;/div&gt;


jgerigmeyer (approved):
I'm approving because this is better than no note, but my leaning would be to remove support for the `anchor` attribute if it looks like it won't eventually land. I don't want to encourage people to use features that have a low percent chance of being implemented natively. @mirisuzanne What do you think?

jgerigmeyer:
@jamesnw Related to this, it looks like the spec has removed `anchor(implicit)`, and added `position-anchor: auto`: https://drafts.csswg.org/css-anchor-position/#changes

mirisuzanne:
> I'm approving because this is better than no note, but my leaning would be to remove support for the `anchor` attribute if it looks like it won't eventually land. I don't want to encourage people to use features that have a low percent chance of being implemented natively.

I agree with @jgerigmeyer

jgerigmeyer (approved):
@jamesnw Feel free to merge if my changes in https://github.com/oddbird/css-anchor-positioning/commit/b076add20c287c233b3bbcd617061ed4da4e2163 look okay.

src/parse.ts
```diff
@@ -138,13 +138,9 @@ function parseAnchorFn(
...
   }
   if (name) {
-    if (isIdentifier(name)) {
-      if (name.name === 'implicit') {
-        name = undefined;
-      } else if (name.name.startsWith('--')) {
-        // Store anchor name
-        anchorName = name.name;
-      }
+    if (isIdentifier(name) && name.name.startsWith('--')) {
```
jamesnw, line 141: I was hesitant to combine these ifs, as it subtly changes the logic, but it ends up not making a difference, so it makes sense to combine.

Issue #276, opened 2024-11-21
Audit polyfill size


jamesnw:
It would be useful to audit the polyfill dist size, with a tool like [rollup-plugin-bundle-stats](https://www.npmjs.com/package/rollup-plugin-bundle-stats). The majority of the code is from csstree, and we may be able to reduce the size by changing how we import things.

jamesnw:
Current state:
<img width="1162" alt="image" src="https://github.com/user-attachments/assets/2a7dddb4-83f2-4bdf-a451-977f6c9401c3">
Bundle Size: 223.72KiB

Moving from `import * as csstree` to more granular imports-
<img width="1166" alt="image" src="https://github.com/user-attachments/assets/7f866d5e-727f-49e2-9a2a-3959014311da">
Bundle Size: 116.87KiB

We might be able to reduce by another 41kb by removing source-map-js- I suspect it is not needed in the compiled bundle.

PR #277, opened 2024-11-22
Reduce built library size

jamesnw:
## Description
This reduces the size of the built library from 223.72KiB to 105.54KiB by:

1. Switching to using more granular imports for csstree. 
2. Removing `source-map-js` from the built library. Source maps are still generated, but the module itself isn't needed.

Pre-merge cleanup:
- [ ] Remove bundle-stats plugin?

## Related Issue(s)
#276 

## Steps to test/reproduce

- Verify that the demo works locally, by building, and then changing the import in index.html to `import polyfill from '/dist/css-anchor-positioning-fn.js';`
- Locally, run `npm run build; open dist/bundle-stats.html` to check the results. (You can compare with the current build on the `prev-stats` branch, which has the bundle-stats plugin installed with no modifications to main.

## Show me

**Before**
image

**After**
image

### DIFF
--- a/src/@types/css-tree.d.ts
+++ b/src/@types/css-tree.d.ts
@@ -3,6 +3,7 @@ declare module 'css-tree/walker' {
 
   export default walk;
 }
+
 declare module 'css-tree/utils' {
   export { clone, List } from 'css-tree';
 }
@@ -12,6 +13,7 @@ declare module 'css-tree/generator' {
 
   export default generate;
 }
+
 declare module 'css-tree/parser' {
   import { parse } from 'css-tree';
 
--- a/src/@types/css-tree.d.ts
+++ b/src/@types/css-tree.d.ts
@@ -1,0 +1,19 @@
+declare module 'css-tree/walker' {
+  import { walk } from 'css-tree';
+
+  export default walk;
+}
+declare module 'css-tree/utils' {
+  export { clone, List } from 'css-tree';
+}
+
+declare module 'css-tree/generator' {
+  import { generate } from 'css-tree';
+
+  export default generate;
+}
+declare module 'css-tree/parser' {
+  import { parse } from 'css-tree';
+
+  export default parse;
+}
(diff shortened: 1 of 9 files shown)


src/@types/css-tree.d.ts
jamesnw, line 1: We could contribute these types upstream, but that would be a larger undertaking, as this is just declaring the pieces we are using.

vite.config.ts
```diff
@@ -37,7 +38,12 @@ export default defineConfig({
         emptyOutDir: !process.env.BUILD_FN,
         target: 'es6',
         sourcemap: true,
+        rollupOptions: {
+          external: [/source-map-js/],
+        },
       },
+  plugins: [bundleStats()],
```
jamesnw, line 45: Because we have multiple outputs, this is just giving us the stats for the last build, but I don't think there's a benefit for auditing the builds separately.

I'm leaning towards leaving this installed, as it can help us keep an eye on the bundle size.

vite.config.ts
```diff
@@ -37,7 +38,12 @@ export default defineConfig({
         emptyOutDir: !process.env.BUILD_FN,
         target: 'es6',
         sourcemap: true,
+        rollupOptions: {
+          external: [/source-map-js/],
```
jamesnw, line 42: Do we need to provide a name in output.globals? There's a warning message emitted-

`No name was provided for external module "source-map-js/lib/source-map-generator.js" in "output.globals". We don't really need to expose it, I don't think?

jgerigmeyer: Yeah, I don't think it really matters. I added it just to silence the warning, since I don't love getting used to "ignore these warnings" if we don't have to.

vite.config.ts
jgerigmeyer: Makes sense. I moved it to `devDependencies` and made it less verbose, but the generated reports are still there.

jgerigmeyer (approved):

Issue #181, opened 2024-05-17
Positioning with `position-area`


jamesnw:
https://drafts.csswg.org/css-anchor-position/#position-area

PR #296, opened 2025-02-17
Support for `position-area`

jamesnw:
## Description
Provides support for `position-area`. 

## Related Issue(s)
#181 

## Steps to test/reproduce
This is in a state where much of the functionality is working. Known issues, which are expected to not be part of an initial release-

- [ ] Shifting content to stay inside its containing block - [Example](https://deploy-preview-296--anchor-polyfill.netlify.app/position-area.html#shifting)
- [ ] `span-all` not centered - [Example](https://deploy-preview-296--anchor-polyfill.netlify.app/position-area.html#span-all-non-centered)
- [ ] `position-area`s as fallback options
- [x] Percentage lengths for margin, etc, adjusted to match new containing block
- [x] Writing direction support

A large caveat is that the positioned element's containing block is generated by creating a grid wrapper element for the positioned element. This means that selectors relying on direct child selectors or sibling selectors will not work. 

## Show me
https://deploy-preview-296--anchor-polyfill.netlify.app/position-area.html

## TODO

- [x] Update README


jamesnw:
@mirisuzanne My initial attempt at positioning the positioned element for a given `position-area` was to closely emulate the spec- setting the inset to the containing block specified by the grid area, and then using `justify-self` and `align-self` to place the element within that containing block. 

However, there is insufficient support for `justify-self` and `align-self` on absolute positioned elements- none in Safari, only 2 months in Firefox, and only a year in Chrome. Is there another way we could do it? 

My current best idea is to create a grid element that is positioned absolutely in the space the new containing block would take up, insert it as the parent to the positioned element, and then we could use `justify-self` and `align-self`.  The major downside I see is that any CSS rules relying on direct child or sibling selectors would fail. Is there a better way to polyfill this?

mirisuzanne:
> My current best idea is to create a grid element that is positioned absolutely in the space the new containing block would take up, insert it as the parent to the positioned element, and then we could use `justify-self` and `align-self`. The major downside I see is that any CSS rules relying on direct child or sibling selectors would fail.

That was my thought process exactly.

I suppose the brute force solution would be creating a sibling positioned grid with an aligned child, and then using the offset from that to re-position the actual element? That could still theoretically cause selector issues, but probably more rare.

I don't know. At a certain point we have to say "it's a polyfill, not a perfect implementation"

jamesnw:
I've pushed an implementation that wraps it in place, and it's working well at https://deploy-preview-296--anchor-polyfill.netlify.app/position-area.html. You can click the unstyled "Toggle Wrapper Visibility" button at the top to see the wrapper. There are still a few issues to work out.

src/parse.ts
```diff
@@ -699,5 +743,68 @@ export async function parseCSS(styleData: StyleData[]) {
...
+      wrapperEl.style.display = 'grid';
+      wrapperEl.style.position = 'absolute';
+      wrapperEl.setAttribute('data-anchor-position-area', positionAreaUUID);
+      targetEl.parentElement?.insertBefore(wrapperEl, targetEl);
+      wrapperEl.appendChild(targetEl);
+      ['top', 'left', 'right', 'bottom'].forEach((prop) => {
+        wrapperEl.style.setProperty(prop, `var(${positionAreaUUID}-${prop})`);
+      });
+
+      // TODO- do these need to be added to inlineStyles as well?
```
jgerigmeyer, line 773: Yes, I think they do. Because `transformCSS` overwrites the inline styles of the element with the modified styles after parsing, any _other_ place where we're calling `setAttribute('style', ...)` after the initial parsing step, we need to _also_ add the same mapping to `inlineStyles` so that `transformCSS` knows to re-add those mappings as well.

src/parse.ts
```diff
@@ -699,5 +743,68 @@ export async function parseCSS(styleData: StyleData[]) {
...
+      // Populate new data for each anchor/target combo
+      validPositions[targetSel] = {
+        ...validPositions[targetSel],
+        positionAreas: [positions],
+        declarations: {
+          ...validPositions[targetSel]?.declarations,
+          [targetProperty]: [
+            ...(validPositions[targetSel]?.declarations?.[
+              targetProperty as InsetProperty
+            ] ?? []),
```
jgerigmeyer, line 801:
```suggestion
            ...(validPositions[targetSel]?.declarations?.[targetProperty] ??
              []),
```

src/polyfill.ts
```diff
@@ -288,34 +290,113 @@ async function applyAnchorPositions(
...
-              reference: anchor,
-              floating: target,
-              strategy: 'absolute',
-            });
-            const resolved = await getPixelValue({
-              targetEl: target,
-              targetProperty: property,
-              anchorRect: rects.reference,
-              anchorSide: anchorValue.anchorSide,
+        if ('positionArea' in anchorValue) {
```
jgerigmeyer, line 300: I think we could also/instead check if `property === 'position-area'`, but maybe this is better.

src/position-area.ts
```diff
@@ -0,0 +1,333 @@
...
+        positionAreas.inline = value[1];
+        positionAreas.block = value[0];
+      }
+      break;
+  }
+  const grid = {
+    block: POSITION_AREA_SPANS[positionAreas.block],
+    inline: POSITION_AREA_SPANS[positionAreas.inline],
+  };
+  const uuid = `--anchor-position-area-${nanoid(12)}`;
```
jgerigmeyer, line 283: Isn't this being overridden by `positionAreaUUID` in `parse.ts`?

src/parse.ts
```diff
@@ -699,5 +743,68 @@ export async function parseCSS(styleData: StyleData[]) {
...
+    const targets: NodeListOf<HTMLElement> =
+      document.querySelectorAll(targetSel);
+    // TODO: Add support for position fallback
+
+    for (const targetEl of targets) {
+      // For every target element, find a valid anchor element
+      const positionAreaUUID = `--position-area-${nanoid(12)}`;
+      const anchorObj = {
+        uuid: `--anchor-${nanoid(12)}`,
+        positionArea: { ...positions.positionArea, uuid: positionAreaUUID },
```
jgerigmeyer, line 759: From what I can tell, `positions.positionArea.uuid` already exists, and was set in `parsePositionAreaValue`. Where is that used?

jgerigmeyer (commented):
@jamesnw This is a great start! I definitely can't say I reviewed every LOC, but did come up with some initial questions. Up to you if it'd be more efficient to talk through these on a call? There's a lot going on in this polyfill, and I get confused quickly 😅

jamesnw:
We are adding a new concept to the polyfill, where we are adding a wrapping element around the target, and then need to apply styles to that element based on the cascade. Currently, we are determining the winner of the cascade by setting a custom property in each selector block with a `position-area` declaration at parse time. Then, at the time we calculate the pixel values for each target, we check to see what the cascaded value is, and apply an attribute to the wrapper. This applies CSS styles that map the custom properties with the pixel values to each inset value.

There is potentially a CSS-only possibility that would allow us to do what we're doing with anchor functions, and let the cascade flow through and determine the values. However, this requires `:has` and `:where` to select the parent of a selector with the same specificity of the selector. 

Support history for :has- https://web-platform-dx.github.io/web-features-explorer/features/has/
Support history for :where: https://web-platform-dx.github.io/web-features-explorer/features/where/

`:has` is the limiting factor, and would move the browser support to:
- Chromium 105
- Firefox 121
- Safari 15.4

We could potentially check for support for :has, and then use the auto-cascade option where possible. Using `@supports selector` would be handy, but that limits us to Safari 14.5 and later. 
https://github.com/web-platform-dx/web-features/blob/main/features/supports.yml.dist#L32-L43

For reference, our current support matrix is:

Firefox 54+ (includes Android)
Chrome 51 - 124 (includes Android)
Edge 79 - 124
Safari 10+ (includes iOS)

schalkneethling:
> `:has` is the limiting factor, and would move the browser support to:
> 
>     * Chromium 105
> 
>     * Firefox 121
> 
>     * Safari 15.4

While the previous support matrix is commendable and mind-blowing (how do you even 😁), I wonder whether the support here is more than good enough, especially if it keeps the code more maintainable and "simpler". What do you all think?

jamesnw:
> (how do you even 😁)

We have been able to largely avoid newer syntax, but much of the early support is theoretical 😀.

> I wonder whether the support here is more than good enough, especially if it keeps the code more maintainable and "simpler". What do you all think?

I think we have a few options. 

1. Use the newer method, and move the support window (with the clarification it's only for `position-area`)
2. Use the older method
3. Support both

I've opened #307 to explore what it would look like to implement this.

jgerigmeyer (approved):
🚀

src/polyfill.ts
```diff
@@ -281,42 +277,142 @@ export const getPixelValue = async ({
   return fallback;
 };
 
+// Use `isPostionAreaDeclaration` instead for type narrowing AST nodes.
```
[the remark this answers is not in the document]
jamesnw: Yeah, that would be a useful pattern, but I think it would be better to do that more holistically in a separate PR.

src/transform.ts
```diff
@@ -36,10 +36,19 @@ export async function transformCSS(
...
-        el.remove();
+        if (!created) {
+          // This is an existing stylesheet, so we replace it.
+          el.insertAdjacentElement('beforebegin', link);
+          // Wait for new stylesheet to be loaded
+          await promise;
+          el.remove();
+        } else {
+          // This is a new stylesheet, so we append it.
+          link.rel = 'stylesheet';
```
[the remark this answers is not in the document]
jamesnw: Existing stylesheets already have a `rel` set, and so that's copied over in line 31 above.

jgerigmeyer (approved):
Looks good! 🎉

jamesnw:
In updating the docs, I discovered that our wrapping element also fixes the margin percentage difference.

Issue #299, opened 2025-03-12
Enable testing of PR versions of Polyfill


jamesnw:
**Is your feature request related to how this polyfill works? Please describe.**
With PRs, there isn't a good way to check that it works on a site besides the demo. The current best way would be to build it from source locally, and then add it to your project.

**Describe the solution you'd like**

We should be able to serve the pre-release version on the Netlify previews by- 

- Building the polyfill dist on netlify PR previews (it currently only has a single compiled source of all JS, including demo code)
- Setting up CORS in the netlify.toml to allow access from other sites

**Describe alternatives you've considered**

We would need to be clear that that this is not for production, and will no longer be available once the PR is merged.

PR #300, opened 2025-03-17
Add built polyfill to Netlify build

jamesnw:
## Description
Build polyfill on Netlify, allow for serving from there

## Related Issue(s)
#299

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -8,7 +8,7 @@ export default defineConfig({
   server: {
     port: 3000,
   },
-  build: process.env.NETLIFY
+  build: process.env.BUILD_DEMO
     ? {}
     : {
         lib: process.env.BUILD_WPT
@@ -35,7 +35,7 @@ export default defineConfig({
                 // the proper extensions will be added
                 fileName: 'css-anchor-positioning',
               },
-        emptyOutDir: !process.env.BUILD_FN,
+        emptyOutDir: false,
         target: 'es6',
         sourcemap: true,
         rollupOptions: {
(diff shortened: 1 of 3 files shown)


vite.config.ts
```diff
@@ -35,7 +35,7 @@ export default defineConfig({
                 // the proper extensions will be added
                 fileName: 'css-anchor-positioning',
               },
-        emptyOutDir: !process.env.BUILD_FN,
```
jamesnw, line 38: If we're building the lib (ie, not the demo build), we don't want to empty the out directory.

jgerigmeyer: This was emptying the out dir once initially (on the `build:dist` run which happens first), and then adding more files to it in the `build:fn` run (which happens second). Not a big deal tho -- maybe we replace this by manually emptying the dir in the `yarn build` step?

jgerigmeyer (approved):
Works for me! I'll let @jerivas speak to any concerns about allowing this in prod -- I can't think of any issues?

@jamesnw Do we want to add some documentation along with this?

vite.config.ts
jamesnw: The demo build doesn't use this setting, so it empties out the dist. Then the regular build and then the fn build run, which include this, so the dist doesn't get overwritten. (I think)

jamesnw:
> @jamesnw Do we want to add some documentation along with this?

Added to CONTRIBUTING.md, since this is for contributors, not for users.

jerivas (approved):

jamesnw:
> I don't see any potential trouble either. My only worry is: if someone uses the script from our Netlify site in a high traffic project, are we going to pay for bandwidth?

Potentially. It looks like we have bandwidth for roughly 6 million requests per month, so it's theoretically not impossible. It's $55 per 100GB over. It would be easy to turn off if it seems like an issue.

PR #302, opened 2025-03-28
Allow multiple position-area declarations for a single selector

jgerigmeyer:
Given:

```css
.target {
  position-area: start;
}

.target {
  position-area: end;
}
```

Before, whichever one we parse last would be applied. Now this stores all of them, and allows the cascade to determine which one should be applied. (Or at least I think so -- some quick functional testing seemed to confirm that this works as expected.)


jamesnw (approved):
Excellent cleanup. Thanks!

PR #309, opened 2025-04-04
Allow anchor-size on margin and inset properties

### LINKED ISSUE #288: Support `anchor-size()` on insets and margins
https://drafts.csswg.org/css-anchor-position-1/#anchor-size-fn

Spec changed in https://github.com/w3c/csswg-drafts/issues/9827

**Additional context**
Added in Chrome 132- https://chromestatus.com/feature/5203950077476864

jamesnw:
## Description
https://drafts.csswg.org/css-anchor-position-1/#anchor-size-fn

Spec changed in https://github.com/w3c/csswg-drafts/issues/9827 to allow `anchor-size` on all properties that are allowed in `@position-try` blocks. This includes some nonsensical ones like `align-self`, which don't take lengths. I went with the "technically correct" tact of allowing anchor size to be used on those, but I alternatively could define a subset of the `@position-try` declarations that actually accept lengths. 

## Related Issue(s)
Fixes #288 

## Steps to test/reproduce
https://deploy-preview-309--anchor-polyfill.netlify.app/#anchor-size-extended

### DIFF
--- a/public/anchor-size-extended.css
+++ b/public/anchor-size-extended.css
@@ -1,0 +1,12 @@
+#my-anchor-anchor-size-extended {
+  anchor-name: --my-anchor-size-extended;
+  resize: both;
+  overflow: hidden;
+}
+
+#my-target-anchor-size-extended {
+  position: absolute;
+  position-anchor: --my-anchor-size-extended;
+  right: anchor-size(width);
+  margin: anchor-size(height);
+}
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -22,6 +22,7 @@ import {
   type AnchorSide,
   type AnchorSize,
   type InsetProperty,
+  isAcceptedPositionTryProp,
   isAnchorSide,
   isAnchorSize,
   isInsetProp,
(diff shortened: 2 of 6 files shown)


src/parse.ts
```diff
@@ -228,8 +232,9 @@ function getAnchorFunctionData(node: CssNode, declaration: Declaration | null) {
       return { changed: true };
     }
     if (
-      isInsetProp(declaration.property) ||
-      isSizingProp(declaration.property)
+      (isAnchorFunction(node) && isInsetProp(declaration.property)) ||
+      (isAnchorSizeFunction(node) &&
+        isAcceptedPositionTryProp(declaration.property))
```
jamesnw, line 237: This splits out the checks, as `anchor-size()` accepts more properties than `anchor()`.

public/anchor-size-extended.css
```diff
@@ -0,0 +1,12 @@
+#my-anchor-anchor-size-extended {
+  anchor-name: --my-anchor-size-extended;
+  resize: both;
```
jgerigmeyer, line 3: This is fun! I approve 🤓

jgerigmeyer (changes requested):
This works well!

> I alternatively could define a subset of the @position-try declarations that actually accept lengths.

My reading of [the spec](https://drafts.csswg.org/css-anchor-position-1/#anchor-size-valid) is actually that `anchor-size` only accepts a subset of the valid `@position-try` declarations that do accept lengths:

> It’s being used in a [sizing property](https://drafts.csswg.org/css-sizing-3/#sizing-property), an [inset property](https://drafts.csswg.org/css-logical-1/#inset-properties), or a [margin property](https://drafts.csswg.org/css-box-4/#margin-properties) on an [absolutely positioned box](https://drafts.csswg.org/css-position-3/#absolute-position). (These are the [accepted @position-try properties](https://drafts.csswg.org/css-anchor-position-1/#accepted-position-try-properties) **that allow lengths**.)

So I'd recommend we split `ACCEPTED_POSITION_TRY_PROPERTIES` from `ACCEPTED_ANCHOR_SIZE_PROPERTIES`.

jgerigmeyer (approved):

Issue #311, opened 2025-04-08
sourcemap error when loading from unpkg


schalkneethling:
When dynamically adding the polyfill:

```js
enablePolyfill.addEventListener("click", async () => {
  const { default: polyfill } = await import(
    "https://unpkg.com/@oddbird/css-anchor-positioning/dist/css-anchor-positioning-fn.js"
  );
  polyfill().then(() => {
    enablePolyfill.textContent = "Polyfill enabled";
  });
});
```

I get the following error in the console related to the sourcemap:

```bash
Uncaught (in promise) TypeError: The specifier “source-map-js/lib/source-map-generator.js” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.
```

I points to the following line at: view-source:https://unpkg.com/@oddbird/css-anchor-positioning@0.5.0/dist/css-anchor-positioning-fn.js

```js
import { SourceMapGenerator as So } from "source-map-js/lib/source-map-generator.js";
```

jamesnw:
I assume this is related to https://github.com/oddbird/css-anchor-positioning/pull/277. It looks like it's also happening for non-dynamic loads from unpkg. Has anyone tested from NPM?

schalkneethling:
> I assume this is related to [#277](https://github.com/oddbird/css-anchor-positioning/pull/277). It looks like it's also happening for non-dynamic loads from unpkg. Has anyone tested from NPM?

Ah yes, looks like here https://github.com/oddbird/css-anchor-positioning/pull/277#discussion_r1854010403

I have not tried with NPM, but I will give it a whirl and report back if nobody beats me to it :)

jamesnw:
I'm also seeing this working in Codepen.

schalkneethling:
> [@jamesnw](https://github.com/jamesnw) [@schalkneethling](https://github.com/schalkneethling) In [#277](https://github.com/oddbird/css-anchor-positioning/pull/277) we externalized the `source-map-js` module (which is imported by `csstree` but not used by this polyfill) to minimize build size, but we didn't remove the import itself -- which is causing the errors in non-dev environments (presumably because the `source-map-js` module exists locally).
>     * https://deploy-preview-312--anchor-polyfill.netlify.app/css-anchor-positioning-fn.js

Tested it with the one above locally and, It Works! 🎉

schalkneethling:
Thank you for the super fast response, folks! 🙌

PR #312, opened 2025-04-08
Remove unused source-map-js import

jgerigmeyer:
One approach to fixing #311 (regression introduced in #277).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
       "devDependencies": {
         "@eslint/js": "^9.24.0",
         "@playwright/test": "1.43.1",
+        "@rollup/plugin-replace": "^6.0.2",
         "@types/async": "^3.2.24",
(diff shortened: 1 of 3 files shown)


jgerigmeyer:
> * This doesn't remove the line from our generated sourcemaps, for instance in `css-anchor-positioning-fn.js.map`, so our sourcemaps may not be correct.

Oh, interesting -- I hadn't considered that. How big of a deal do we think that is? 🤔 

https://github.com/rollup/plugins/discussions/1798

> * There is still a call to SourceMapGenerator in the bundled plugin. This is from css-tree, and looks like it would only be called if we called `generate({sourceMap: true})`, which we are not. It would be nice if we could remove that codepath from our bundled css-tree, but I don't think that's necessary.

Yeah, true. The codepath is 75 LOC -- not sure how far we want to go down this rabbit hole!

Issue #317, opened 2025-04-12
[BUG] Fallbacks not triggered with the viewport as the containing block


thejackshelton:
Hey there! I'm currently working on the successor to our current popover component:

https://qwikui.com/docs/headless/popover/

Ideally we'd like to use the css anchor API polyfill here rather than floating ui by itself with a limited api

---

Problem:

When adding any fallbacks, the polyfill does not go into a fallback position.

Minimal reproduction:

https://codepen.io/Jack-Shelton/pen/emYqMqa

**Expected behavior**

The floating UI flip middleware should execute, similar to when you go to this codepen on Chrome

**Browsers tested on**
Include OS, Browser, Browser Version, and device (if mobile).

MacOS 15.1 Sequoia, Chrome 135, Mac M4
MacOS 15.1 Sequoia, Safari 18.3, Mac M4
MacOS 15.1 Sequoia, Firefox 135, Mac M4
iOS, Safari 18.3, Safari iPhone 15

schalkneethling:
@thejackshelton You may already have seen this, and I am unsure that the limitations are impacting you here, but just in case:
https://github.com/oddbird/css-anchor-positioning?tab=readme-ov-file#limitations as there is some mention about `flip-*`

thejackshelton:
> [@thejackshelton](https://github.com/thejackshelton) You may already have seen this, and I am unsure that the limitations are impacting you here, but just in case: https://github.com/oddbird/css-anchor-positioning?tab=readme-ov-file#limitations as there is some mention about `flip-*`

https://anchor-positioning.oddbird.net/#position-try-tactics-combined

yes, you can see that the behavior here seems to work. however when copying this example verbatim I was unable to get the same behavior in 0.5.1

jamesnw:
Hi! Thanks for reporting, and I'm happy to help make this work for Qwik. This appears to be a bug, and my current thought is that it's due to the scroll container being the viewport. I'm troubleshooting and hope to find the solution quickly.

Side note, we don't use the Floating UI `flip` middleware- we use Floating UI to aid in finding the positions of elements on the page, but are recreating the flip logic so that it matches the CSS spec.

schalkneethling:
> we use Floating UI to aid in finding the positions of elements on the page

Fascinating! I had no idea you all were using it under the hood 👀

thejackshelton:
Hey @jamesnw! Appreciate the clarification! 

[...]

Issue #279, opened 2024-11-22
[BUG] Position-try-fallbacks demo breaks at certain screen widths


jgerigmeyer:
Still investigating, but at certain screen widths "Target 2" reverts to its original (un-polyfilled) location in the position-try-fallbacks demo.

https://anchor-positioning.oddbird.net/#position-try-tactics

**Browsers tested on**

FF 132

jamesnw:
I think this example is broken because we are currently using cascaded values where we shouldn't be. In this case, the scroll container has a `height: 400px`. This gets converted to `--height-uuid: 400px`, and then when we are generating the fallbacks, we are checking for `--height-uuid`. All the fallbacks are ending up with a `height: 400px` when they shouldn't.

We only need to set a height when there's a height set directly on the element, not if an ancestor sets a height. We'll need to find a way to differentiate if a custom property is set directly on an element vs inherited.

I took a quick look through the accepted position try properties, and I don't think any of them are inherited.

jamesnw:
The demo no longer works in nonpolyfilled browsers either- I don't think our initial work on this was correct.

PR #318, opened 2025-04-14
Alternate overflow options

jamesnw:
## Description
Overflow checking was not working as expected for `position-try-fallbacks` with the viewport as the containing block. 

## Related Issue(s)
#317 

## Steps to test/reproduce
- Verify [contained example ](https://deploy-preview-318--anchor-polyfill.netlify.app/#position-try)still works (Note that the Try Tactics example is broken currently- see #279)
- Verify the [viewport example](https://codepen.io/jamessw/pen/OPPLgrm) works in Codepen

### DIFF
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -441,12 +441,13 @@ async function checkOverflow(target: HTMLElement, offsetParent: HTMLElement) {
       y: target.offsetTop,
       platform: platformWithCache,
       rects,
-      elements: { floating: target },
+      elements: {
+        floating: target,
+        reference: offsetParent,
+      },
       strategy: 'absolute',
     } as unknown as MiddlewareState,
     {
-      boundary: offsetParent,
-      rootBoundary: 'document',
       padding: getMargins(target),
     },
   );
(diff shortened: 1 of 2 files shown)


src/polyfill.ts
```diff
@@ -441,12 +441,13 @@ async function checkOverflow(target: HTMLElement, offsetParent: HTMLElement) {
...
-      elements: { floating: target },
+      elements: {
+        floating: target,
+        reference: offsetParent,
+      },
       strategy: 'absolute',
     } as unknown as MiddlewareState,
     {
-      boundary: offsetParent,
-      rootBoundary: 'document',
```
jamesnw, line 449: As context, if I recall correctly, I was being overly explicit in this call, but it turns out these options are not always correct, and the defaults are actually what we want.

jamesnw:
> Probably unrelated, but I notice in my Chromium browser (no polyfill) that the last successful fallback "sticks" even if scrolled to allow a prior fallback to no longer overflow. Is that a bug, or a change in the spec?

I've noticed this as well. I opened https://github.com/oddbird/css-anchor-positioning/issues/325 to track this.

PR #323, opened 2025-04-18
Make relative css imports absolute to allow loading from blobs

### LINKED ISSUE #274: [BUG] The polyfill causes @font-face rules to be ignored
**Describe the bug**
When @font-face rules are in CSS sources that are parsed by the polyfill, it seems like the navigator does not load them again after the CSS have been re-injected in the blob source. Please not that the bug is present only with self-hosted fonts.

**To Reproduce**
Example to reproduce here: https://projects.c-mh.fr/anchor-polyfill-font-bug/

**Expected behavior**
The loaded fonts needs to be still displayed.

**Browsers tested on**
Chrome (by forcing polyfill execution) and Firefox.

**Possible workaround**
For now, I have declared my fonts in a separate CSS file that won't be parsed by the polyfill, but in my case, it's not an ideal solution.

### LINKED ISSUE #314: [BUG] CSS imports not working with 0.5.1
**Describe the bug**
I have a CSS file that imports another CSS file. When the polyfill is enabled the imported styles stop working. I rolled back to 0.5.0 and it started working again.

**To Reproduce**
```css
@import url('./typography.css');

p {
  color: black;
}
```

typography.css: 
```css
p {
  color: red;
}
```

The second stylesheet will load but not be applied.

**Browsers tested on**
Firefox 137 on Mac

jamesnw:
## Description

This provides partial support for rules in CSS Imports. 

Currently, when the polyfill re-writes a stylesheet and loads it as a blob, any relative URLs in an `@import` statement will not be able to be loaded. 

This change rewrites the URL to be absolute, based on `document.baseURI`, so that styles from imported stylesheets are applied.

This change does NOT parse imported stylesheets for anchor positioning declarations.

It also does not convert urls in imported stylesheets, but this shouldn't be a problem, as they should still be able to load content from a relative URL.

## Alternate Approach
This is an exploration to one approach for handling the broken relative URLs. The alternate approach is to inject the styles into the document directly, rather than as links.

## Related Issue(s)
Fixes #274, Fixes #314

## Steps to test/reproduce
https://deploy-preview-323--anchor-polyfill.netlify.app/#imports


jamesnw:
We are initially going to try the approach in #324, but may revisit if testing by users reveals any reasons to favor this instead.

PR #324, opened 2025-04-18
Replace blob approach with style elements

jamesnw:
## Description
An alternate approach to #323 that injects all styles into the head, even if they were originally links.

Forked from css-imports to verify the example works.

Style and Link elements support different atttributes, and we'll want to revisit https://github.com/oddbird/css-anchor-positioning/pull/262 to make sure there aren't impacts there. I think the most important one is `media`, which is supported.

## Related Issue(s)
#314 #274 

## Steps to test/reproduce
https://deploy-preview-324--anchor-polyfill.netlify.app/


jamesnw:
@mirisuzanne I'm curious if you have thoughts on the best approach between this PR and #323. 

## The problem
We currently replace `<link rel=stylesheet src="url">` with `<link rel=stylesheet src="blob:...>`. This works, except that relative URLs in `@imports` or declarations that support a `url()` can't be loaded. 

## Solution A
(#323)
Find all `url()` calls in all declarations (and strings in `@import`) and resolve them to make them absolute.

Pros: Smaller change in behavior
Cons: Adding another parse pass of any stylesheets that have been changed

## Solution B
(#324, this one)
Replace all `<link rel=stylesheet src="url">` with `<style>[css]</style>` in the document head.

Pros: no need to reparse, may prevent other bugs that have not yet been encountered
Cons: Fairly large change to the document, adding a lot of content to the document directly.

Both seem to work, but are there other issues we should take into consideration, or any of these items that you see as  particularly impactful?

mirisuzanne:
I don't have any particular input here, it seems like you're asking the right questions.

src/transform.ts
```diff
@@ -20,36 +34,31 @@ export async function transformCSS(
         // Handle inline stylesheets
         el.innerHTML = css;
       } else if (el instanceof HTMLLinkElement) {
-        // Create new link
-        const blob = new Blob([css], { type: 'text/css' });
-        const url = URL.createObjectURL(blob);
-        const link = document.createElement('link');
+        // Replace link elements with style elements
```
jgerigmeyer, line 37: I think we should add a brief note (and/or link to the issue/pr) explaining why we're doing this.

jgerigmeyer (approved):
👏

jgerigmeyer:
@jamesnw Do we have an issue for tracking whether we want to follow `@imports` to polyfill them as well? If not let's add one.

jamesnw:
> @jamesnw Do we have an issue for tracking whether we want to follow `@imports` to polyfill them as well? If not let's add one.

Good call. I created #326.

jgerigmeyer:
@bartveneman Great -- thanks for testing it out and letting us know! We'll likely do a full release today or early next week.

PR #336, opened 2025-06-12
Report invalid selectors

### LINKED ISSUE #335: [BUG] crash: TypeError: can't access property "map", t.children is undefined
**Describe the bug**
```
Uncaught (in promise) TypeError: can't access property "map", t.children is undefined
    cn css-anchor-positioning:4400
    enter css-anchor-positioning:5473
    Pn css-anchor-positioning:602
    h css-anchor-positioning:690
    m css-anchor-positioning:693
    reduce css-anchor-positioning:3530
    Rn css-anchor-positioning:646
    h css-anchor-positioning:691
    Rn css-anchor-positioning:648
    h css-anchor-positioning:691
    m css-anchor-positioning:693
    reduce css-anchor-positioning:3530
    Rn css-anchor-positioning:646
    h css-anchor-positioning:691
    u css-anchor-positioning:718
    bp css-anchor-positioning:5469
    Pp css-anchor-positioning:5728
    R css-anchor-positioning:28
    R css-anchor-positioning:14
    Pp css-anchor-positioning:5724
    Ps css-anchor-positioning:6331
    o css-anchor-positioning:17
    promise callback*l css-anchor-positioning:27
    R css-anchor-positioning:28
    R css-anchor-positioning:14
    Ps css-anchor-positioning:6325
    Hp css-anchor-positioning:6336
    n browser:7
    setupOnce browser:7
    it dom.ts:95
    Hp css-anchor-positioning:6335
    bo css-anchor-positioning:13
    <anonymous> css-anchor-positioning:6339
```

**To Reproduce**

```js
// fix of Firefox & safari's anchor position
if (!("anchorName" in document.documentElement.style))
	import("https://unpkg.com/@oddbird/css-anchor-positioning")
```

**Expected behavior**
Include relevant parts of the [spec](https://drafts.csswg.org/css-anchor-position-1/), and a clear and concise description of what you expected to happen.
no crash.

**Screenshots**

Image

**Browsers tested on**
Include OS, Browser, Browser Version, and device (if mobile).
Firefox Dev 140.0b7, windows11

**Additional context**
```js
// fix of Firefox & safari's anchor position
if (!("anchorName" in document.documentElement.style))
	import("https://unpkg.com/@oddbird/css-anchor-positioning")
```
it freeze the page for few sec and then throw an error.

jamesnw:
## Description
Invalid selectors, for instance `.w-[10]`, crash the polyfill. This PR throws a more specific error to guide the developer in fixing it. It does not attempt to ignore or fix the error. 

There may be other places where invalid CSS will break the polyfill's assumption of validity, and we could use a similar approach as they are reported.

## Related Issue(s)
Closes #335 

## Steps to test/reproduce
Locally, add a rule with a selector like `.w-[10]` that is invalid, and see that the error is thrown.

Or, view console at https://codepen.io/jamessw/pen/GgJdEeM

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,17 @@ style properties at runtime.
 If you are using inline styles to set anchor-related properties and the polyfill
 isn't working, verify that the inline styles are actually showing up in the DOM.
 
+### Invalid CSS
+
+Some types of invalid CSS will cause the polyfill to throw an error. In these
+cases, the polyfill will report any parse errors encountered in the console as
+warnings. This will be followed by the error thrown by the polyfill.
+
+The polyfill can't determine which parse error caused the polyfill error, but
+please resolve any reported parse errors before opening a bug. We also recommend
+using a CSS linter like [Stylelint](https://stylelint.io/) or
+[@eslint/css](https://github.com/eslint/css).
+
 ## Sponsor OddBird's OSS Work
 
 At OddBird, we love contributing to the languages & tools developers rely on.
--- a/src/cascade.ts
+++ b/src/cascade.ts
@@ -50,7 +50,7 @@ function shiftUnsupportedProperties(node: CssNode, block?: Block) {
 export function cascadeCSS(styleData: StyleData[]) {
   for (const styleObj of styleData) {
     let changed = false;
-    const ast = getAST(styleObj.css);
+    const ast = getAST(styleObj.css, true);
     walk(ast, {
       visit: 'Declaration',
       enter(node) {
--- a/src/polyfill.ts
+++ b/src/polyfill.ts
@@ -32,7 +32,7 @@ import {
   type SizingProperty,
 } from './syntax.js';
 import { transformCSS } from './transform.js';
-import { cssParseErrors } from './utils.js';
+import { reportParseErrorsOnFailure, resetParseErrors } from './utils.js';
 
 const platformWithCache = { ...platform, _c: new Map() };
 
@@ -591,8 +591,17 @@ export async function polyfill(
 
   // fetch CSS from stylesheet and inline style
   let styleData = await fetchCSS(options.elements, options.excludeInlineStyles);
+
   let rules: AnchorPositions = {};
   let inlineStyles: Map<HTMLElement, Record<string, string>> | undefined;
+
+  // Reset the CSS parse errors in case the polyfill is run multiple times, and
+  // at the beginning in case a previous run failed.
+  resetParseErrors();
+
+  // If the polyfill fails during the steps in the try catch, it is likely due
+  // to invalid CSS, so report the CSS parse errors. Subsequent errors are less
+  // likely to be caused by parse errors.
   try {
     // pre parse CSS styles that we need to cascade
     const cascadeCausedChanges = cascadeCSS(styleData);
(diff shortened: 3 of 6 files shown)


jgerigmeyer (commented):
I'm okay merging this, but it feels like it might be better to more generally expose parsing errors from CSSTree rather than catching this one specific selector case. I think we could use [onParseError](https://github.com/csstree/csstree/blob/master/docs/parsing.md#onparseerror) in our single call to CSSTree `parse`, e.g.:

```js
    onParseError: (err) => {
      throw new Error(err.formattedMessage);
    },
```

Or if we don't want to stop execution, in case the invalid CSS doesn't require the polyfill:

```js
    onParseError: (err) => console.error(err.formattedMessage),
```

@jamesnw What do you think?

jamesnw:
I think you're right that a more global option would be helpful.

I'm still leaning towards throwing an error and stopping execution. Catching the error in `onParseError` doesn't fix or skip that node, so the polyfill still executes (and throws an additional error) on that node.

I updated this PR with this. I included the original error as the cause, and also added a note to make it clear that the polyfill did not complete.

jamesnw:
Hmm. I don't think the polyfill should take the role of a linter, and it could be confusing if the polyfill is reporting CSS parse errors that are unrelated to the polyfill working. It also is confusing, as you point out, for the polyfill to not work if there are unrelated CSS parse errors that don't break the polyfill.

In the ideal situation, the polyfill would only surface parse errors if it results in the polyfill not running. There are two ways of doing that- first would be identifying parse errors that do break the polyfill, and catch and throw those. But as you point out, that's going to be pretty context specific.

The other way would be storing all the parse errors, and then wrapping the entire polyfill in a try catch. If it throws, then reporting the parse errors. I don't like that idea either.

My current leaning is going back to the original solution- catching individual errors in places where they will break the polyfill, and throwing with better guidance on the fix. Thoughts?

jgerigmeyer:
> The other way would be storing all the parse errors, and then wrapping the entire polyfill in a try catch. If it throws, then reporting the parse errors. I don't like that idea either.

I guess this approach doesn't really bother me, and it seems much more straightforward than trying to anticipate places in our code where malformed CSS (resulting in an unparsed `Raw` node) will trigger polyfill errors. It seems like as a user, if the polyfill can't run for whatever reason, I would love to see a list of any invalid CSS. 🤔

I don't feel strongly though, so if you think logging from `onParseError` is heavy-handed, then we can just handle this one case now and see if other bug reports arise.

jamesnw:
@jgerigmeyer I haven't fixed the tests yet, but I'm curious on your take on this. I think I have it reporting only parse errors in the user's stylesheets, and only when the polyfill fails.

I store all the parse errors during the cascade step, since that's the initial pass through the CSS. Otherwise, we'd be reporting on errors parsing stylesheets that were changed by the polyfill.

I only wrapped the cascade and parse steps in the try catch, as they are the steps that parse CSS. While it's possible that a syntax error would cause the polyfill to fail in later steps, I think it's more likely that the syntax error would be unrelated, and likely to be a red herring.

 It does report all parse errors, and doesn't attempt to identify which parse error caused the polyfill to fail.

jgerigmeyer (commented):
@jamesnw This looks good to me!

Issue #191, opened 2024-05-17
Anchors and targets in shadow dom


jamesnw:
Related: https://github.com/w3c/csswg-drafts/issues/9408

Issue #170, opened 2024-04-22
Can I load this in a custom-element, so it can access shadow-dom?


herrKlein:
Hi, really nice polyfill. 

I would like the example which keeps the container in the viewport. I have 2 questions.
Does it work with shadowdom. I would like to use it for a custom dropdown component where the popup has to stay inside the window.

Second question is, I see the example constantly updating when it overlaps the border of the scrollable content. Wouldn't that make it slow. Is that a bug?

![Screenshot 2024-04-22 at 11 30 53](https://github.com/oddbird/css-anchor-positioning/assets/6085660/6ee6dd17-2fee-4e40-ba5a-46b39a43d137)

jgerigmeyer:
> Does it work with shadowdom. I would like to use it for a custom dropdown component where the popup has to stay inside the window.

That's a great question, and my answer is honestly "I'm not sure." We paused development on this polyfill to allow the spec to stabilize (see comment [here](https://github.com/oddbird/css-anchor-positioning/issues/167#issuecomment-2048472235)), and I don't think we've tested with custom components yet. If you have time to test that out, please report back!

> Second question is, I see the example constantly updating when it overlaps the border of the scrollable content. Wouldn't that make it slow. Is that a bug?

Yes, that certainly looks like a bug to me. Would you be willing to open a new issue?

haoliangwu:
I've confirmed it on my personal project, it seems this polyfill will not cover the custom elements(with using shadow dom), all the polyfills works as expected on light dom. If you use custom elements without shadow dom, I think it should also work.

PR #353, opened 2025-10-01
Work with anchor and target inside same shadow root

wkillerud:
## Description

Thanks for making it possible to use anchor positioning while supporting older browsers ❤️ 

I ran into one of the limitations of the current polyfill while writing a web component where both the anchor and target are inside the same shadow root. The reliance on `document.querySelectorAll` to find the elements to measure makes it so the current version can't work in shadow roots.

Open for other approaches here, but the gist of it is to pass in a shadow root if there is one to run `querySelectorAll` on that instead of `document`. I went for an approach binding the call to the component instance, but it could also be an extra `root` option to the function call that defaults to `document` or something like that.

## Related Issue(s)

- #170 
- #191 (somewhat related, though in this case both anchor and target are inside the same shadow root)

## Steps to test/reproduce

```
npm run serve
```

Then visit http://localhost:3000/shadow-root.html

## Show me

https://github.com/user-attachments/assets/5f6c0e75-4426-4c5e-bd5f-a4c776f1ec2a


jamesnw:
Thanks for this contribution! I'd definitely love to improve how the polyfill works with the shadow DOM. I don't necessarily think a PR needs to solve the entire problem to be helpful or merged, but am thinking as I type about how to make sure this is headed in the correct direction.

From the [spec](https://drafts.csswg.org/css-anchor-position-1/#name)- 

> Positioned elements in [shadow trees](https://dom.spec.whatwg.org/#concept-shadow-tree) can reference [anchor names](https://drafts.csswg.org/css-anchor-position-1/#anchor-name) defined in “higher” trees. Currently, they cannot reference anchor names defined in “lower” shadow trees, though.

Would this be possible with this approach? 

The other portion of the spec related to shadow DOM is about using `::part`, which I think is somewhat orthogonal to this PR.

> I went for an approach binding the call to the component instance, but it could also be an extra `root` option to the function call that defaults to `document` or something like that.

Hmm, this is interesting. I see two uses for a `root` option-

1. As an additive property- "here are additional roots to look at, in addition to `document`". This would address the use case in this PR. It would perhaps allow for tethering to an anchor in a higher tree, although we may need to consider the higher/lower relationships somehow
2. As a scoping property- "All my positioned elements are in my site's navbar, so my root is that element". There might be some performance gain from limiting the search, although that may be minimal in practice. 

I think both use cases would be possible by allowing an array, something like `{root: [document, shadowRoot1, shadowRoot2]}`. My sense is that a `root` option would be more flexible than binding the call.

wkillerud:
> From the [spec](https://drafts.csswg.org/css-anchor-position-1/#name)-
>
> > Positioned elements in [shadow trees](https://dom.spec.whatwg.org/#concept-shadow-tree) can reference [anchor names](https://drafts.csswg.org/css-anchor-position-1/#anchor-name) defined in “higher” trees. Currently, they cannot reference anchor names defined in “lower” shadow trees, though.
>
> Would this be possible with this approach?

It should be possible with some adjustments. I double-checked that it's possible for a [child component to navigate up through a parent shadow root and into light DOM](https://codepen.io/wkillerud/pen/JoGRMKW?editors=1011). We can use the presence of a `host` to know if we're entering a new shadow tree vs light DOM, for scoping.

We'll need an element locator that does this type of navigation up and into light DOM, querying along the way. Once it reaches light DOM it could skip straight to `document` unless we want to have that scoping property you mention with the `root` option.

> The other portion of the spec related to shadow DOM is about using ::part, which I think is somewhat orthogonal to this PR.

Have to admit I'm not that familiar with `::part` and how that would play out with the polyfill 😅 The selectors seem to be in the parent shadow tree or light DOM, but the anchor element itself is inside the shadow root, so measuring it for positioning would be tricky?

> My sense is that a root option would be more flexible than binding the call.

That makes sense, let's do that instead then.

wkillerud:
> Positioned elements in [shadow trees](https://dom.spec.whatwg.org/#concept-shadow-tree) can reference [anchor names](https://drafts.csswg.org/css-anchor-position-1/#anchor-name) defined in “higher” trees.

Do you know if this is implemented anywhere yet? The spec update was pretty recent, so maybe not. I'm trying to [set up an example](https://github.com/wkillerud/css-anchor-positioning/pull/1/files) and control that it works in Chrome (checked Stable and Canary). I might very well be doing something wrong, would love to not "fly blind" 😅

wkillerud:
The last test run failed with a diff above a threshold. Set up a devcontainer to see if it was an OS specific thing, but the tests pass there too. Maybe OK with a retry?

> I'd be content with opening a separate issue to support that, and knowing that this change doesn't make that change harder.

Sounds good.

I think the `root` option is a good first step to add support. If a root has a `host` property it's a shadow tree and would need some extra steps.

shadow-root.html
jamesnw, line 1: I'm fine with having a separate HTML doc for this (the index.html page is quite long), but I do think we need a way to get to this page from there. The position-area examples are set up with a single example on index.html, and then a link from that example to the other examples, but this only has one example. 

Perhaps add a `section.demo-item` on `index.html`, but without a `.demo-elements` section? I'm open to other suggestions.

This could become a page with more examples of shadow dom- perhaps rename it to `shadow-dom.html`? 

Also, you'll need to add this file to `vite.config.ts`, in `rollupOptions.input`.

src/dom.ts
```diff
@@ -140,18 +141,19 @@ function getContainerScrollPosition(element: HTMLElement) {
...
+  selector: Selector,
+  options: { root: AnchorPositioningRoot[] },
+) {
   const { elementPart, pseudoElementPart } = selector;
   const result: (HTMLElement | PseudoElement)[] = [];
   const isBefore = pseudoElementPart === '::before';
   const isAfter = pseudoElementPart === '::after';
 
-  // Current we only support `::before` and `::after` pseudo-elements.
+  // Currently we only support `::before` and `::after` pseudo-elements.
```
jamesnw, line 153: Thanks!

src/fetch.ts
```diff
@@ -96,11 +98,10 @@ function fetchInlineStyles(elements?: HTMLElement[]) {
 }
 
 export async function fetchCSS(
-  elements?: HTMLElement[],
-  excludeInlineStyles?: boolean,
+  options: AnchorPositioningPolyfillOptions,
```
jamesnw, line 101: Hmm, this makes me wonder if we should have an internal type for resolved options, where the root is required? It might be nice, but not required.

src/polyfill.ts
```diff
@@ -553,6 +555,12 @@ export interface AnchorPositioningPolyfillOptions {
   // An array of explicitly targeted elements to polyfill
   elements?: HTMLElement[];
 
+  /**
+   * Set the root elements that are queried when looking for anchors and targets.
+   * @default document
+   */
+  root?: AnchorPositioningRoot[];
```
jamesnw, line 562: Thoughts on this being `roots` to reflect that it's an array?

src/polyfill.ts
```diff
@@ -578,6 +586,10 @@ function normalizePolyfillOptions(
     options.elements = undefined;
   }
 
+  if (!Array.isArray(options.root)) {
```
jamesnw, line 589: Should we also check if people are passing `{root: []}` for some reason? I think that should default to `[document]` as well, instead of putting the polyfill into a state that will error.

We could also handle `{root: shadowRoot}` (without an array) potentially, but I'm less convinced that we should do this.

jamesnw (commented):
This is looking great! I have some nits and questions. The tests did seem to pass on re-run.

@jgerigmeyer Do you have any objection to adding the `.devcontainer`?

One more area that would be useful to update would be the README.md. Could you document the `root(s)` option under Configuration, and also update the line about shadow-dom  (L153) to reflect your changes?

Thanks so much for this contribution!

src/polyfill.ts
wkillerud: I think with `roots` plural it makes sense to just have the array for now, but I could add it if you'd like.

jamesnw: I agree, let's expect an array.

wkillerud:
> I'd probably prefer to remove the .devcontainer/ files, unless we add more documentation around them.

I'll remove it then, it's quick to set up should anyone want it for themselves 👍

PR #403, opened 2026-06-04
Support CSS anchor positioning in constructed stylesheets (`adoptedStyleSheets`)

### LINKED ISSUE #228: The polyfill won’t work with `CSSStyleSheet()` or `CSSStyleDeclaration()`

jpzwarte:
## Summary

Extends the polyfill to handle CSS anchor positioning defined in [constructed stylesheets](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet) that are adopted into shadow roots via `adoptedStyleSheets`. This is the typical pattern for styles in custom elements using libraries like Lit.

## Architecture

Support is delivered as a **separate entrypoint** (`/shadow`) so it is opt-in and doesn't affect the bundle size of the default entrypoint. Users who need this feature import the shadow entrypoint in addition to (or instead of) the default one:

```html
<!-- Must be loaded before any custom element's connectedCallback runs -->
<script type="module" src="/dist/css-anchor-positioning-shadow.js"></script>
```

Or via npm:

```js
import 'css-anchor-positioning/shadow';
```

The entrypoint calls `installConstructedStylesheetPatches()` from the new `src/shadow.ts` module, which installs the required patches at module-evaluation time.

### New files

- **`src/shadow.ts`** — contains all patching logic for constructed stylesheets
- **`src/index-shadow.ts`** — dedicated entrypoint; calls `installConstructedStylesheetPatches()` and re-exports `polyfill`

#### 1. `CSSStyleSheet.prototype.replaceSync`

The native `replaceSync` is patched to capture the original CSS text in a `WeakMap<CSSStyleSheet, string>` (`adoptedSheetText`) before passing it to the native implementation. This ensures the polyfill can later re-parse the untransformed source text via `fetchCSS`, even after the sheet has been updated with transformed CSS.

#### 2. `ShadowRoot.prototype.adoptedStyleSheets` setter

The setter is patched to call `patchHostConnectedCallback(shadowRoot)` whenever non-empty sheets are adopted. This wraps the host element's `connectedCallback` so that after the original callback runs (and the shadow DOM is fully populated), `polyfill({ roots: [shadowRoot] })` is invoked. Two timing cases are handled:

- **Sheets adopted before the host connects** — the wrapped `connectedCallback` fires on the next DOM connection.
- **Sheets adopted while the host is already connected** (e.g. from inside `connectedCallback`) — a `queueMicrotask` defers the polyfill run until after the current callback stack finishes, so that `innerHTML` and other DOM setup has completed.

A `WeakSet<HTMLElement>` (`patchedHosts`) ensures each host's callback is only wrapped once, even if multiple sheets are adopted.

#### `src/utils.ts`

- `StyleData.el` is now optional (can be absent for stylesheet-backed entries)
- `StyleData.sheet?: CSSStyleSheet` — new optional field linking a `StyleData` entry to its constructed stylesheet
- `originalReplaceSync` — saved reference to the native `replaceSync` for write-back
- `adoptedSheetText` — `WeakMap` storing captured source CSS per sheet
- `captureAdoptedStylesheetText`, `getAdoptedStylesheetText`, `writeAdoptedStylesheet` — helpers used by `shadow.ts`, `fetch.ts`, and `transform.ts`

#### `src/fetch.ts`

New `fetchAdoptedStyleSheets` helper collects adopted `CSSStyleSheet` objects from each polyfill root, building `StyleData` entries (with a `sheet` reference and no `el`) using the captured source text (or serialized `cssRules` as a fallback). These are appended to the sources passed to `fetchLinkedStylesheets`. Skipped when `options.elements` is provided, since that opts into element-scoped polyfilling only.

#### `src/transform.ts`

- When a `StyleData` entry has a `sheet`, writes transformed CSS back into it via `writeAdoptedStylesheet` (which calls `originalReplaceSync` directly, bypassing the patch so the captured source text is not overwritten).
- Guards `el?.tagName`, `el?.hasAttribute`, and `el?.removeAttribute` calls with optional chaining, since `el` is now optional.

#### `vite.config.ts` / `package.json`

- New `build:shadow` script (`cross-env BUILD_SHADOW=1 vite build`) builds the shadow entrypoint
- New `./shadow` package export pointing to `dist/css-anchor-positioning-shadow.{js,umd.cjs,d.ts}`

### Demo (`shadow-dom.html`)

Updated to demonstrate the `adoptedStyleSheets` pattern:

- When `#apply-polyfill` is present in the URL hash, the shadow entrypoint is imported first, then custom element definitions are registered (so patches are in place before any `connectedCallback` runs).
- A new `<anchor-adopted-styles>` component demonstrates the full pattern using constructed stylesheets.

### Tests

A new e2e test (`tests/e2e/shadow-dom.test.ts`) verifies that anchored elements inside a shadow root styled via `adoptedStyleSheets` are correctly positioned after the polyfill runs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,12 +33,20 @@
       "types": "./dist/index-fn.d.ts",
       "import": "./dist/css-anchor-positioning-fn.js",
       "require": "./dist/css-anchor-positioning-fn.umd.cjs"
+    },
+    "./shadow": {
+      "types": "./dist/index-shadow.d.ts",
+      "import": "./dist/css-anchor-positioning-shadow.js",
+      "require": "./dist/css-anchor-positioning-shadow.umd.cjs"
     }
   },
   "typesVersions": {
     "*": {
       "fn": [
         "./dist/index-fn.d.ts"
+      ],
+      "shadow": [
+        "./dist/index-shadow.d.ts"
       ]
     }
   },
@@ -52,6 +60,7 @@
     "build": "run-s build:demo build:dist build:fn",
     "build:dist": "vite build",
     "build:fn": "cross-env BUILD_FN=1 vite build",
+    "build:shadow": "cross-env BUILD_SHADOW=1 vite build",
     "build:wpt": "cross-env BUILD_WPT=1 vite build",
     "build:demo": "cross-env BUILD_DEMO=1 vite build",
     "preview": "vite preview",
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "package.json"
   ],
   "scripts": {
-    "build": "run-s build:demo build:dist build:fn",
+    "build": "run-s build:demo build:dist build:fn build:shadow",
     "build:dist": "vite build",
     "build:fn": "cross-env BUILD_FN=1 vite build",
     "build:shadow": "cross-env BUILD_SHADOW=1 vite build",
--- a/shadow-dom.html
+++ b/shadow-dom.html
@@ -23,7 +23,8 @@
     </script>
     <link rel="stylesheet" href="/demo.css" />
     <style>
-      anchor-web-component {
+      anchor-web-component,
+      anchor-adopted-styles {
         width: 100%;
       }
     </style>
@@ -37,13 +38,14 @@
 
       if (!SUPPORTS_ANCHOR_POSITIONING) {
         btn.addEventListener('click', () => {
-          document
-            .querySelector('anchor-web-component')
-            .applyPolyfill()
-            .then(() => {
-              btn.innerText = 'Polyfill Applied';
-              btn.setAttribute('disabled', '');
-            });
+          const roots = [
+            document.querySelector('anchor-web-component').shadowRoot,
+            document.querySelector('anchor-adopted-styles').shadowRoot,
+          ];
+          polyfill({ roots }).then(() => {
+            btn.innerText = 'Polyfill Applied';
+            btn.setAttribute('disabled', '');
+          });
         });
       } else {
         btn.innerText = 'No Polyfill Needed';
@@ -53,13 +55,36 @@
         );
       }
 
-      class AnchorWebComponent extends HTMLElement {
-        /* This would typically be run in connectedCallback() */
-        applyPolyfill() {
-          return polyfill({ roots: [this.shadowRoot] });
+      class AnchorWebComponent extends HTMLElement {}
+      customElements.define('anchor-web-component', AnchorWebComponent);
+
+      class AnchorAdoptedStyles extends HTMLElement {
+        connectedCallback() {
+          this.attachShadow({ mode: 'open' });
+
+          const sheet = new CSSStyleSheet();
+          sheet.replaceSync(`
+            .anchor {
+              anchor-name: --adopted-anchor;
+            }
+            .target {
+              position: absolute;
+              right: anchor(--adopted-anchor right, 50px);
+              top: anchor(--adopted-anchor bottom);
+            }
+          `);
+          this.shadowRoot.adoptedStyleSheets = [sheet];
+
+          this.shadowRoot.innerHTML = `
+            <link rel="stylesheet" href="/demo.css" />
+            <div style="position: relative">
+              <div class="target">Target</div>
+              <div class="anchor">Anchor</div>
+            </div>
+          `;
         }
       }
-      customElements.define('anchor-web-component', AnchorWebComponent);
+      customElements.define('anchor-adopted-styles', AnchorAdoptedStyles);
     </script>
     <link rel="icon" href="/favicon.svg" type="image/svg+xml" />
     <script src="https://unpkg.com/prismjs@v1.x/components/prism-core.min.js"></script>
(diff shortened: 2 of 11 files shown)


src/transform.ts
```diff
@@ -84,7 +89,7 @@ export function transformCSS(
       }
     }
     // Remove no-longer-needed data-attribute
-    if (cleanup && el.hasAttribute('data-has-inline-styles')) {
+    if (cleanup && el?.hasAttribute('data-has-inline-styles')) {
```
jamesnw, line 92: I don't think this change is needed?

src/polyfill.ts
```diff
@@ -32,10 +32,110 @@ import {
...
+  const host = shadowRoot.host as CustomElementHost;
+  if (patchedHosts.has(host)) {
+    return;
+  }
+  patchedHosts.add(host);
+
+  const originalConnectedCallback = host.connectedCallback;
+  host.connectedCallback = function (this: CustomElementHost) {
+    originalConnectedCallback?.call(this);
+    void polyfill({ roots: [shadowRoot] });
```
jamesnw: Should this pass in the options from the parent polyfill call?

jamesnw (commented):
This is great, and seems to work well! I wanted to get back with high level feedback and questions, before jumping into a line by line review.

I think the general method makes sense, and I think I mostly have questions about how it's exposed.

Could we avoid patching for users who don't need it? One option would be to move the patching to a separate entry point, like `index-shadow.ts`, that would call `installConstructedStylesheetPatches` before applying the polyfill? Perhaps the patching is exposed as an entry point separate from the polyfill, and then the user imports both.

Another item to consider at some point is documentation, but that can wait until the interface above is settled.

Thanks for this work!

src/transform.ts
jpzwarte: So in the `fetchAdoptedStyleSheets()` function, there is no `el`. I've adjusted `StyleData` to make `el` optional.

src/polyfill.ts
jpzwarte: There is no "parent polyfill call", right? Or should the first call to `polyfill()` cache the options passed in a global variable etc. and then reuse them here?

jpzwarte:
> This is great, and seems to work well! I wanted to get back with high level feedback and questions, before jumping into a line by line review.

I just checked in the iPadOS Simulator and it works there as well.
 
> I think the general method makes sense, and I think I mostly have questions about how it's exposed.
> 
> Could we avoid patching for users who don't need it? One option would be to move the patching to a separate entry point, like `index-shadow.ts`, that would call `installConstructedStylesheetPatches` before applying the polyfill? Perhaps the patching is exposed as an entry point separate from the polyfill, and then the user imports both.

Done, there is now a new entrypoint that calls `installConstructedStylesheetPatches()` and still exports the `polyfill` function. I've updated the `shadow-dom.html` accordingly.
 
> Another item to consider at some point is documentation, but that can wait until the interface above is settled.

I've updated `shadow-dom.html` as best i can.

package.json
```diff
@@ -52,6 +60,7 @@
     "build": "run-s build:demo build:dist build:fn",
```
jamesnw:
```suggestion
    "build": "run-s build:demo build:dist build:fn build:shadow",
```

shadow-dom.html
```diff
@@ -23,43 +23,84 @@
...
+            roots: [document.querySelector('anchor-web-component').shadowRoot]
+          });
+
+          btn.innerText = 'Polyfill Applied';
+          btn.setAttribute('disabled', '');
+        } else {
+          btn.addEventListener('click', () => {
+            location.hash = 'apply-polyfill';
+            location.reload();
+          });
```
jamesnw, line 93:
```suggestion
          });
          defineCustomElements();
```

This way, it loads on non-supported browsers without being polyfilled.

shadow-dom.html
jamesnw: I think this could mention/link to the example below.

src/polyfill.ts
jamesnw: Good point. There is a way to set global variables to pass options to the polyfill so users wouldn't be completely without control here. I think I would recommend leaving it as is, and seeing if any issues arise. If people run into this, we can figure out a way of handling it.

src/utils.ts
```diff
@@ -63,11 +63,54 @@ export function getDeclarationValue(node: DeclarationWithValue) {
...
   created?: boolean; // Whether the element is created by the polyfill
+  // The constructed stylesheet this data came from, when the styles were
+  // adopted via `adoptedStyleSheets` rather than a `<style>`/`<link>` element.
+  sheet?: CSSStyleSheet;
+}
+
+// Reference to the native `CSSStyleSheet.prototype.replaceSync` so that the
+// polyfill can write transformed CSS back into a constructed stylesheet without
+// re-triggering the patched version (which would re-capture the text).
+export const originalReplaceSync = CSSStyleSheet.prototype.replaceSync;
```
jamesnw, line 79: Would we have any savings by moving these to shadow.ts? I assume this would get tree shaken for the non-shadow builds but I haven't checked.

jamesnw (changes requested):
This is looking great and I think it's really close. Would you also be willing to update the Readme?  I'd also be willing to do that if you'd prefer - you've already done a ton here. If you'd prefer I do it, I can do it in a follow up PR, or I think you can enable "Allow edits from maintainers" if it isn't already. 

Also, if you pull in main, there is a fix that would allow this to be tested in the deploy preview. 

Thanks!

src/shadow.ts
```diff
@@ -0,0 +1,88 @@
...
+  const host = shadowRoot.host as CustomElementHost;
+  if (patchedHosts.has(host)) {
+    return;
+  }
+  patchedHosts.add(host);
+
+  const originalConnectedCallback = host.connectedCallback;
+  host.connectedCallback = function (this: CustomElementHost) {
+    originalConnectedCallback?.call(this);
+    void polyfill({ roots: [shadowRoot] });
```
jamesnw, line 27: I don't think we need e2e tests for this, or even a demo, but have you confirmed that both this and the "host already connected" path below both trigger as expected?

src/utils.ts
jpzwarte: `transform.ts` calls `writeAdoptedStylesheet` which in turn calls `originalReplaceSync`. So i don't think it makes sense to move the latter 2 to `shadow.ts`, right?

src/shadow.ts
jpzwarte: I have (i just had copilot generate a demo+test for it), but i can see it adds little additional value as a demo. I'll remove it.

shadow-dom.html
jpzwarte: Done

jpzwarte:
> Would you also be willing to update the Readme?

Done!

jamesnw (approved):
This looks good! I'll merge and publish an alpha, then plan to give it a week for testing. Thanks so much for your contribution!

jamesnw:
Linted and tested locally, since the CI wasn't running.

PR #407, opened 2026-06-10
Move installConstructedStylesheetPatches to -fn entry

jamesnw:
Removes the `index-shadow` entry, and moves the installConstructedStylesheetPatches function to the index-fn entry point. 

I'm also wondering about changing `installConstructedStylesheetPatches` to `patchAndPolyfillConstructedStylesheets()`? It would be nice to be clear that CEs using Custom Stylesheets will be automatically polyfilled, but it's a bit verbose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,11 +66,11 @@ If your custom elements use [constructed stylesheets](https://developer.mozilla.
 ```html
 <script type="module">
   if (!('anchorName' in document.documentElement.style)) {
-    const { default: polyfill, patchAndPolyfillConstructedStylesheets } =
+    const { patchAndPolyfillConstructedStylesheets } =
       await import('https://unpkg.com/@oddbird/css-anchor-positioning/dist/css-anchor-positioning-fn.js');
     patchAndPolyfillConstructedStylesheets();
-    await polyfill();
-    // Define your custom elements after the polyfill has loaded
+    // Define your custom elements after the patch has applied the polyfill.
+    // You don't need to explicitly call polyfill().
   }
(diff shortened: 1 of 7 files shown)


jpzwarte:
Changes look good to me. Name change is a good idea as well!

README.md
```diff
@@ -59,31 +61,38 @@ been applied.
...
 ```html
 <script type="module">
   if (!('anchorName' in document.documentElement.style)) {
-    await import('https://unpkg.com/@oddbird/css-anchor-positioning/dist/css-anchor-positioning-shadow.js');
-    // Define your custom elements after the entrypoint has loaded
+    const { default: polyfill, patchAndPolyfillConstructedStylesheets } =
+      await import('https://unpkg.com/@oddbird/css-anchor-positioning/dist/css-anchor-positioning-fn.js');
+    patchAndPolyfillConstructedStylesheets();
+    await polyfill();
+    // Define your custom elements after the polyfill has loaded
```
jgerigmeyer: I'm not 100% clear on this comment. In the shadow-dom.html example, it looks like the custom elements are defined after `patchAndPolyfillConstructedStylesheets()` but before `polyfill()`.

jgerigmeyer (approved):

README.md
jamesnw: The `polyfill()` on shadow-dom.html applies the polyfill to the other example, not the custom elements. I'll update this to make it clearer.

jgerigmeyer (approved):

PR #409, opened 2026-06-11
Add tooltip custom element with position-anchor support on the `:host`

### LINKED ISSUE #408: [BUG] `position-anchor` pointing from the `:host` of a custom element to another element is broken
The following HTML snippet doesn't work correctly with the polyfill:

```html
<button style="anchor-name: --button">Button</button>
<x-tooltip style="position-anchor: --button">
  <!-- more CSS anchor positioning within the shadow root of the custom element -->
</x-tooltip>
```

You can see some polyfill changes to the CSS of the custom element, but the actual positioning is broken.

jpzwarte:
Fixes #408

## What this fixes

When `position-anchor` is set on a **custom element host** (i.e. via a `:host` rule in an adopted stylesheet), the polyfill failed to position the element. Three separate issues combined to cause this:

1. **`:host` targets were never matched.** `querySelectorAll(':host')` returns nothing when called from within a shadow root, so the polyfill never recognised the custom element as a target. Fixed in `src/dom.ts` by resolving `:host`/`:host(...)` selectors directly against the shadow root's host element.

2. **The anchor lived outside the polyfilled roots.** The anchor element (e.g. a `<div>` in the light DOM) was not in any of the shadow roots passed to `polyfill()`, so `getAnchorEl` couldn't find it. Fixed in `src/parse.ts` by also searching the target element's own root node when looking up its anchor.

3. **Inline styles set via typed CSSOM were silently dropped.** Setting `el.style.anchorName = ...` in a browser without native anchor-positioning support discards the unknown declaration, leaving no `style` attribute for the polyfill to read. Fixed in the demo (`shadow-dom.html`) by writing the declarations as raw attribute strings via `setAttribute('style', ...)`, which browsers preserve literally.

## New demo

A `<position-anchor-on-host>` custom element is added to `shadow-dom.html`. Its `connectedCallback` wires up the anchor link at runtime:

```html
<div id="anchor" class="anchor">Anchor</div>
<position-anchor-on-host for="anchor">Target</position-anchor-on-host>
<script>
class PositionAnchorOnHost extends HTMLElement {
  connectedCallback() {
    const targetId = this.getAttribute("for"),
      target = targetId && this.getRootNode().getElementById(targetId);

    this.setAttribute("style", "position-anchor: --position-anchor-on-host");
    target?.setAttribute("style", "anchor-name: --position-anchor-on-host");

    this.attachShadow({ mode: "open" });

    const sheet = new CSSStyleSheet();
    sheet.replaceSync(`
      :host {
        top: anchor(top);
        left: anchor(center);
        position: absolute;
        translate: -50% -100%;
      }
    `);
    this.shadowRoot.adoptedStyleSheets = [sheet];
    this.shadowRoot.innerHTML = "<slot></slot>";
  }
}
customElements.define("position-anchor-on-host", PositionAnchorOnHost);
</script>
```

### DIFF
--- a/shadow-dom.html
+++ b/shadow-dom.html
@@ -72,8 +72,14 @@
             // properties, so no `style` attribute is produced and the polyfill
             // never sees them. Setting the literal attribute string preserves
             // the declarations for the polyfill to read.
-            this.setAttribute('style', 'position-anchor: --position-anchor-on-host');
-            target?.setAttribute('style', 'anchor-name: --position-anchor-on-host');
+            this.setAttribute(
+              'style',
+              'position-anchor: --position-anchor-on-host',
+            );
+            target?.setAttribute(
+              'style',
+              'anchor-name: --position-anchor-on-host',
+            );
 
             this.attachShadow({ mode: 'open' });
 
@@ -329,8 +335,8 @@ class AnchorAdoptedStyles extends HTMLElement {
         </p>
         <p>
           This demo uses <code>position-anchor</code> on a custom element host
-          (<code>&lt;position-anchor-on-host&gt;</code>), with <code>anchor()</code> in
-          the shadow root's <code>:host</code> styles.
+          (<code>&lt;position-anchor-on-host&gt;</code>), with
+          <code>anchor()</code> in the shadow root's <code>:host</code> styles.
         </p>
       </div>
 
@@ -376,7 +382,6 @@ class PositionAnchorOnHost extends HTMLElement {
 }
 customElements.define("position-anchor-on-host", PositionAnchorOnHost);
 &lt;/script&gt;
-</code></pre>
 </code></pre>
     </section>
     <section id="sponsor">
--- a/shadow-dom.html
+++ b/shadow-dom.html
@@ -60,6 +60,52 @@
           }
         }
         customElements.define('anchor-adopted-styles', AnchorAdoptedStyles);
+
+        class Tooltip extends HTMLElement {
+          connectedCallback() {
+            const anchorName = `--x-tooltip-${Math.random().toString(36).slice(2)}`;
+            const targetId = this.getAttribute('for');
+            const target = targetId && this.getRootNode().getElementById(targetId);
+            // Write the anchor properties as raw `style` attribute strings.
+            // In browsers without native anchor-positioning support, the typed
+            // CSSOM (`el.style.anchorName = ...`) silently drops these unknown
+            // properties, so no `style` attribute is produced and the polyfill
+            // never sees them. Setting the literal attribute string preserves
+            // the declarations for the polyfill to read.
+            if (target) {
+              target.setAttribute(
+                'style',
+                `anchor-name: ${anchorName}; ${target.getAttribute('style') ?? ''}`,
+              );
+            }
+            this.setAttribute(
+              'style',
+              `position-anchor: ${anchorName}; ${this.getAttribute('style') ?? ''}`,
+            );
+
+            this.attachShadow({ mode: 'open' });
+
+            const sheet = new CSSStyleSheet();
+            sheet.replaceSync(`
+              :host {
+                position: absolute;
+                top: calc(anchor(top) - 4px);
+                left: anchor(center);
+                background: #222;
+                color: #fff;
+                padding: 4px 8px;
+                border-radius: 4px;
+                font-size: 0.8em;
+                white-space: nowrap;
+                pointer-events: none;
+                translate: -50% -100%;
+              }
+            `);
+            this.shadowRoot.adoptedStyleSheets = [sheet];
+            this.shadowRoot.innerHTML = '<slot></slot>';
+          }
+        }
+        customElements.define('x-tooltip', Tooltip);
       }
 
       const btn = document.getElementById('apply-polyfill');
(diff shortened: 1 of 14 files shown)


shadow-dom.html
```diff
@@ -60,6 +60,51 @@
...
+            // never sees them. Setting the literal attribute string preserves
+            // the declarations for the polyfill to read.
+            this.setAttribute(
+              'style',
+              'position-anchor: --position-anchor-on-host',
+            );
+            target?.setAttribute(
+              'style',
+              'anchor-name: --position-anchor-on-host',
+            );
```
jamesnw: Is this necessary for the functionality? I think you can test the same functionality by setting the style directly on the light dom elements, and it makes it a bit easier to understand.

shadow-dom.html
```diff
@@ -271,6 +317,71 @@ <h2>
...
+
+    const sheet = new CSSStyleSheet();
+    sheet.replaceSync(`
+      :host {
+        --element-color: var(--target, var(--outer-anchored));
+        background: var(--element-color);
+        border: thin solid var(--border);
+        border-radius: var(--radius-1);
+        color: white;
+        font-weight: bold;
```
jamesnw:
```suggestion
```

I think we can exclude the styling CSS from the code block to simplify the example (but keep it in the actual custom element definition).

src/dom.ts
```diff
@@ -258,11 +258,42 @@ export const getOffsetParent = async (el: HTMLElement) => {
...
+ * `:host(<compound-selector>)` selector. The `:host` pseudo-class cannot be
+ * matched via `querySelectorAll()` from within the shadow root, so we resolve
+ * it manually against the host element.
+ */
+function hostMatchesSelector(host: HTMLElement, selector: string): boolean {
+  const trimmed = selector.trim();
+  if (trimmed === ':host') {
+    return true;
+  }
+  const match = /^:host\(\s*(.+?)\s*\)$/.exec(trimmed);
```
jamesnw, line 275: I think this works, but could we parse it as CSS rather than using regex? I think that might allow us to handle selector lists like `:host(.blue, .red)`

jamesnw (changes requested):
I think this is adding a bug. If the `position-anchor` is defined in the Constructed Stylesheet, on `:host`, and the `anchor-name` is defined in the light dom, browsers do not find the anchor-name. With this change, the polyfill does find the anchor.

```
:host{
  position-anchor: --a;
}
```

I'm not entirely certain of what is correct per the spec here. My understanding of the spec is that the target should not be able to find the anchor. The anchor has to be named in the same tree where it is used. 

If that's the case, I'd like to figure out the more general pattern we could apply, so that the polyfill works for your example, without overeagerly exposing anchor names.

shadow-dom.html
jpzwarte: Fixed

shadow-dom.html
jpzwarte: Done

src/dom.ts
jpzwarte: Done, but i'm not sure exactly what you meant with "parse it as CSS".

jpzwarte:
> I think this is adding a bug. If the `position-anchor` is defined in the Constructed Stylesheet, on `:host`, and the `anchor-name` is defined in the light dom, browsers do not find the anchor-name. With this change, the polyfill does find the anchor.

I think you're right. I've tested all 3 major browsers and they show the same behavior. I'll investigate how to fix the polyfill.

jpzwarte:
Done. It now only positions the element when `position-anchor` is an inline style, not `:host {}` scoped.

src/dom.ts
jamesnw: I was thinking using `csstree` to parse the selector string, and there is a `getSelectors` helper in `utils`, but I'm not sure if that matches the use case here. Essentially, convert the selector into the AST, and then work on that. That avoids some potential edge cases, like an escaped comma in a selector.

src/dom.ts
```diff
@@ -258,11 +258,42 @@ export const getOffsetParent = async (el: HTMLElement) => {
...
+}
+
 export const querySelectorAllRoots = (
   roots: AnchorPositioningRoot[],
   selector: string,
 ): HTMLElement[] => {
-  return roots.flatMap(
-    (e) => [...e.querySelectorAll(selector)] as HTMLElement[],
-  );
+  return roots.flatMap((root) => {
```
jamesnw, line 290: Worst feedback ever coming- I can't quite put my finger on it, but this seems not quite right. I'm wondering if the need to check inline vs internal styles elsewhere is papering over a logic issue here. 

Does this function allow leakage between roots? If so, then we have similar issues with styles leaking for other uses of `querySelectorAllRoots`. 

Or maybe it's addressing tree scoping in the `validatedForPositioning` function to match the [spec](https://drafts.csswg.org/css-anchor-position-1/#target)?

src/parse.ts
```diff
@@ -263,13 +263,29 @@ function getAnchorFunctionData(node: CssNode, declaration: Declaration | null) {
...
+/**
+ * Whether the element declares `position-anchor` in its own inline `style`
+ * attribute. This is used to distinguish a `position-anchor` that is tree-scoped
+ * to the element's own tree (e.g. an inline style on a shadow host, in the light
+ * DOM) from one inherited via a shadow-scoped `:host` rule. Only the former may
+ * reference an anchor in the host's outer tree.
+ */
+function hasInlinePositionAnchor(el: HTMLElement): boolean {
+  const style = el.getAttribute('style');
+  return style ? /(?:^|;)\s*position-anchor\s*:/i.test(style) : false;
```
jamesnw, line 275: I think this usage of regex is ok, and you wouldn't need to parse this as CSS.

jamesnw (changes requested):
A few more questions, I could be completely wrong- thanks for your continued work on this!

jpzwarte:
I'm closing this PR in favor of #415

PR #410, opened 2026-06-12
Add `inset: auto` to a polyfilled popover

### LINKED ISSUE #330: [BUG] Popover with a position-area not positioned correctly
**Describe the bug**
Popovers are not positioned correctly, and require an additional`inset: auto` declaration to be positioned correctly. 

**To Reproduce**

1. Go to https://codepen.io/jamessw/pen/yyyKYwE?editors=1100
2. In Chromium, both popovers are positioned correctly
3. In polyfilled browsers, `inset: auto` must be set on the popover for it to be positioned correctly, as in the top button/popover.

**Expected behavior**
Users shouldn't need to add additional styles for the polyfill to work. We should either apply it, or find a way to not need to apply it.

**Browsers tested on**
Safari, Firefox

jpzwarte:
Closes #330 

Fixes positioning of polyfilled popovers that use `position-area` by explicitly setting `inset: auto` on the target element inside the wrapper. Without this, the browser's default `inset` values on `[popover]` elements conflict with the wrapper-based positioning approach.

Also adds a demo example to `index.html` showing a popover anchored via `position-area` instead of `anchor()`.

### DIFF
--- a/public/anchor-popover-position-area.css
+++ b/public/anchor-popover-position-area.css
@@ -7,6 +7,7 @@
 #my-target-popover-position-area {
   position: absolute;
   margin: 0;
+  inset: 10px;
   position-anchor: --my-anchor-popover-position-area;
   position-area: end;
 }
--- a/public/anchor-popover-position-area.css
+++ b/public/anchor-popover-position-area.css
@@ -5,9 +5,7 @@
 }
 
 #my-target-popover-position-area {
-  position: absolute;
-  margin: 0;
-  inset: 10px;
+  inset: 10px 5%;
   position-anchor: --my-anchor-popover-position-area;
   position-area: end;
 }
--- a/public/anchor-popover-position-area.css
+++ b/public/anchor-popover-position-area.css
@@ -1,0 +1,12 @@
+#my-anchor-popover-position-area {
+  position: absolute;
+  left: 200px;
+  anchor-name: --my-anchor-popover-position-area;
+}
+
+#my-target-popover-position-area {
+  position: absolute;
+  margin: 0;
+  position-anchor: --my-anchor-popover-position-area;
+  position-area: end;
+}
--- a/src/position-area.ts
+++ b/src/position-area.ts
@@ -554,8 +554,13 @@ export function wrapperForPositionedElement(
     wrapperEl.style.pointerEvents = 'none';
     targetEl.style.pointerEvents = originalPointerEvents;
 
-    // The wrapper sets the inset values, so the target should not have them.
-    targetEl.style.inset = 'auto';
+    // A popover in the top layer receives `inset: 0` from the UA stylesheet,
+    // which would conflict with the wrapper-based positioning. Reset it so the
+    // wrapper's inset values take effect. Non-popover targets already default to
+    // `inset: auto`, so this is only needed for popovers.
+    if (targetEl.hasAttribute('popover')) {
+      targetEl.style.inset = 'auto';
+    }
 
     ['top', 'left', 'right', 'bottom'].forEach((prop) => {
       wrapperEl.style.setProperty(prop, `var(--pa-value-${prop})`);
--- a/src/position-area.ts
+++ b/src/position-area.ts
@@ -562,8 +562,7 @@ export function wrapperForPositionedElement(
     // off the target so the wrapper drives positioning, and re-apply any author
     // offset as padding on the wrapper. Because the wrapper represents the
     // `position-area` cell and the target is aligned within it, padding shrinks
-    // the cell and offsets the target inward — matching `inset` semantics. The
-    // UA's `inset: 0` resolves to zero padding, so it is a no-op. Non-popover
+    // the cell and offsets the target inward — matching `inset` semantics. Non-popover
     // targets stay in this wrapper's containing block, so their own `inset`
     // already works and is left untouched.
     if (targetEl.hasAttribute('popover')) {
--- a/src/position-area.ts
+++ b/src/position-area.ts
@@ -567,6 +567,7 @@ export function wrapperForPositionedElement(
     // already works and is left untouched.
     if (targetEl.hasAttribute('popover')) {
       const targetStyles = getComputedStyle(targetEl);
+     // Logical insets can also be retrieved via their physical equivalents, so we don't need to loop through those as well.
       (['top', 'right', 'bottom', 'left'] as const).forEach((side) => {
         const value = targetStyles[side];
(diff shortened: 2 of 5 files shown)


jamesnw (changes requested):
Thanks for taking this on! I think this has 2 issues-

1. You've applied it to position-area uses, not popover. Not all popovers use position-area, and not all position-area uses are popover. 
2. This overwrites existing insets on the element, but they need to be maintained.

Maybe the algorithm is "if the positioned element has the popover attribute, get the computed value of each inset. If it is `0`, set to auto". (I'm less certain of tthe second step)

jpzwarte:
> Thanks for taking this on! I think this has 2 issues-

Does it make sense to have both `position-area` *and* `inset`?

I've added an if around setting `inset: auto` that checks if it's a popover.

jamesnw:
> > Thanks for taking this on! I think this has 2 issues-
> 
> Does it make sense to have both `position-area` _and_ `inset`?
> 
> I've added an if around setting `inset: auto` that checks if it's a popover.



> > Thanks for taking this on! I think this has 2 issues-
> 
> Does it make sense to have both `position-area` _and_ `inset`?

Yeah, it is a legitimate use case- setting an inset allows you to position the content within the containing block generated by `position-area`, for instance if you want to offset it slightly from the anchor.
> 
> I've added an if around setting `inset: auto` that checks if it's a popover.

This still misses the case where it is a popover not using position-area. For instance, the demo above the one you added is setting margin:0 to accomplish the same thing. If inset: auto is applied to that, the margin can be removed.

jpzwarte:
> Yeah, it is a legitimate use case- setting an inset allows you to position the content within the containing block generated by `position-area`, for instance if you want to offset it slightly from the anchor.

I'll work on this; thanks.

> > I've added an if around setting `inset: auto` that checks if it's a popover.
> 
> This still misses the case where it is a popover not using position-area. For instance, the demo above the one you added is setting margin:0 to accomplish the same thing. If inset: auto is applied to that, the margin can be removed.

I think you're wrong here :) When i remove `margin: 0`, the native behavior matches the polyfilled behavior in the "Anchoring a popover" demo. So afaict, `margin: 0` is needed even when you're not using the polyfill. I think this is due to the `margin: auto` user agent style.

jpzwarte:
> I'll work on this; thanks.

I've got it working! It checkes if the target is a popover and has a non-zero `inset` value, it removes the inset and sets it on the wrapper element.

public/anchor-popover-position-area.css
```diff
@@ -0,0 +1,13 @@
+#my-anchor-popover-position-area {
+  position: absolute;
+  left: 200px;
+  anchor-name: --my-anchor-popover-position-area;
+}
+
+#my-target-popover-position-area {
+  position: absolute;
```
jamesnw: With #413  merged in, I think you can remove `position: absolute` here, and use the popover's fixed from the UA.

public/anchor-popover-position-area.css
```diff
@@ -0,0 +1,13 @@
...
+  margin: 0;
+  inset: 10px;
```
jamesnw:
```suggestion
  inset: 10px 5%;
```
Just to show handling multiple values and percent.

src/position-area.ts
```diff
@@ -554,6 +554,29 @@ export function wrapperForPositionedElement(
...
+    // containing block becomes the viewport rather than this wrapper. That means
+    // any non-`auto` `inset` on the popover — the UA stylesheet's `inset: 0`, or
+    // an author value such as `inset: 10px` — pins it to the viewport instead of
+    // letting it take its place inside the wrapper's grid cell. Strip the inset
+    // off the target so the wrapper drives positioning, and re-apply any author
+    // offset as padding on the wrapper. Because the wrapper represents the
+    // `position-area` cell and the target is aligned within it, padding shrinks
+    // the cell and offsets the target inward — matching `inset` semantics. The
+    // UA's `inset: 0` resolves to zero padding, so it is a no-op. Non-popover
+    // targets stay in this wrapper's containing block, so their own `inset`
```
jamesnw:
```suggestion
    // the cell and offsets the target inward — matching `inset` semantics. Non-popover
    // targets stay in this wrapper's containing block, so their own `inset`
```

I don't think this is a useful comment

src/position-area.ts
```diff
@@ -554,6 +554,29 @@ export function wrapperForPositionedElement(
...
+    // already works and is left untouched.
+    if (targetEl.hasAttribute('popover')) {
+      const targetStyles = getComputedStyle(targetEl);
+      (['top', 'right', 'bottom', 'left'] as const).forEach((side) => {
```
jamesnw, line 572:
```suggestion
     // Logical insets can also be retrieved via their physical equivalents, so we don't need to loop through those as well.
      (['top', 'right', 'bottom', 'left'] as const).forEach((side) => {
```

src/position-area.ts
```diff
@@ -554,6 +554,29 @@ export function wrapperForPositionedElement(
...
+        const value = targetStyles[side];
+        if (value && value !== 'auto' && parseFloat(value) !== 0) {
+          wrapperEl.style.setProperty(`padding-${side}`, value);
```
jamesnw, line 575: This is a clever way of addressing this. I think it should be documented on the README, as well as on the position-area.html page.

public/anchor-popover-position-area.css
```diff
@@ -0,0 +1,13 @@
+#my-anchor-popover-position-area {
+  position: absolute;
+  left: 200px;
+  anchor-name: --my-anchor-popover-position-area;
+}
+
+#my-target-popover-position-area {
+  position: absolute;
+  margin: 0;
```
jamesnw: Can you check without this again? I'm not seeing it make an impact in polyfilled or non-polyfilled browsers.

jamesnw (changes requested):
This is close! Just a few small changes.

public/anchor-popover-position-area.css
jpzwarte: Done

public/anchor-popover-position-area.css
jpzwarte: Done

public/anchor-popover-position-area.css
jpzwarte: Me neither. Weird though because i could have sworn i saw it make a difference last week...

src/position-area.ts
jpzwarte: Done

jamesnw (approved):
This looks good! Thanks for your work on this!

PR #412, opened 2026-06-12
Add `positionAreaContainingBlock` option

### LINKED ISSUE #411: [BUG] Wrapping the element in `<polyfill-position-area>` may break existing CSS
As it says in the README.md:

```
- `position-area` is polyfilled by adding a wrapping element around the target,
  which adds a few differences:
  - This breaks selectors that rely on a direct relationship with the target,
    for instance `~ target`, `+ target`, `> target` or using `:nth` selectors.
  - Overflow alignment is not applied for a target that overflows its
    inset-modified containing block but would still fit within its original
    containing block. In other words, a polyfilled target may be placed in a
    `position-area` grid section outside its containing block, where the
    implementation would move the target inside the containing block.
```

I don't have an alternative solution at the moment, but i just wanted an issue to track it.

jpzwarte:
Closes #411

Adds a `positionAreaContainingBlock` option (default: `true`) to control how `position-area` is polyfilled.

- **`true` (default):** The target is wrapped with a `<polyfill-position-area>` element that approximates the containing block created by `position-area`. Alignment is applied via `justify-self`/`align-self` on the target within that wrapper.
- **`false`:** No wrapper element is inserted. Instead, the polyfill resolves the `position-area` alignment to concrete inset values and applies them directly on the target. This avoids breaking selectors that depend on a direct DOM relationship (e.g. `> target`, `+ target`, `:nth-child()`), but comes with trade-offs around inset overrides, self-alignment, and overflow alignment (documented in the README and `position-area.html`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ value of `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS`.
     window.ANCHOR_POSITIONING_POLYFILL_OPTIONS = {
       elements: undefined,
       excludeInlineStyles: false,
+      positionAreaContainingBlock: true,
       roots: [document],
       useAnimationFrame: false,
     };
@@ -122,6 +123,7 @@ an argument.
     polyfill({
       elements: undefined,
       excludeInlineStyles: false,
+      positionAreaContainingBlock: true,
       roots: [document],
       useAnimationFrame: false,
     });
@@ -148,6 +150,19 @@ that have eligible inline styles, regardless of whether the `elements` option is
 defined. When set to `true`, elements with eligible inline styles listed in the
 `elements` option will still be polyfilled, but no other elements in the
 document will be implicitly polyfilled.
+
+### positionAreaContainingBlock
+
+type: `boolean`, default: `true`
+
+By default, `position-area` is polyfilled by wrapping the target with a
+`<polyfill-position-area>` element that approximates the containing block
+created by `position-area`. When set to `false`, no wrapper element is added,
+and the polyfill computes and applies inset values on the target itself
+instead. This avoids breaking selectors that rely on a direct relationship
+with the target (for instance `> target` or `:nth-child()`), but comes with
+its own trade-offs. See [Limitations](#limitations) for the differences
+between the two approaches.
 
 ### roots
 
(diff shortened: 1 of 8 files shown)


jamesnw:
Thanks for taking a look at this! I haven't done a full review, but wanted to note the questions I will be considering when reviewing. 

1. What are the differences between having a containing block and not? (See https://www.oddbird.net/2025/02/25/anchor-position-area/#a-new-containing-block)
2. Can we programmatically discover when a containing block will have an impact? 
3. How do we want to expose this? Is a global setting for the polyfill sufficient? Or maybe some way of disabling the containing block for certain elements? A `data-anchor-position-polyfill-disable-cb` attribute, or `--anchor-position-polyfill-disable` custom property? I don't like the idea of leaving cleanup for when the polyfill can be removed, but control could be useful. I also want to avoid something that people opt out of because it fixes one problem, and then get surprised by the other differences.

jpzwarte:
So my reasoning for this PR is as follows:

1. It's just easier to position an element using `position-area` with a fallback:

```css
  position-area: top;
  position-try-fallbacks: bottom;
```

2. As far as i can tell, not having a wrapper element doesn't impact *positioning* the element. It *does* impact sizing the element. So without the wrapper element, `padding-left: 50%` differs from the native implementation.
3. The wrapper element can break CSS and JavaScript (for manual slot assignments for example in custom elements).

So i kind of see this as a trade-off: you lose the relative sizes a containing block gives you, but you gain not having a wrapper element and having to take that into account in a web component based design system.

Since in our design system we so far always use CSS Anchor Positioning hand-in-hand with popover, it would like to have that trade-off.

To answer you questions:
1. See above
2. Good question; the first thing i can think of is detecting if percentages are used for padding etc.
3. I like your idea of signaling the polyfill to not use a wrapper element, but i'm not sure asking every developer to be aware of this is a good idea.

jpzwarte:
As an experiment i've added an "auto" setting. This checks if the target has any styling that is affected by the containing block. It seems to work: 
- "span-left top" is wrapped (uses `padding-right: 50%`)
- "center left" isn't

jamesnw:
I think the criteria for how to expose this is going with the least surprising method, with the clearest resolution. It seems like there's room for surprise no matter what-

1. Wrapped by default > My selector doesn't work.
2. Unwrapped by default > My padding is wrong.
3. Wrapped only if needed (auto) by default > Why is this different?

My thought is going with #1, and keeping it wrapped by default. This means it isn't a breaking change for polyfill users. It also means the resolution is clearer- they inspect the DOM, see a wrapper from the polyfill, and can investigate how to disable it.

With #2, the resolution is less clear- they see the padding is different, but wouldn't have a clear signal that this is something they need to set in the polyfill.

I do think all 3 options would be useful to have.

jpzwarte:
Closing this in favor of #428

Issue #350, opened 2025-09-14
[BUG] Scrolling the viewport moves popovertarget downwards if inside a container with position: sticky


matthiasott:
**Describe the bug**
I am using anchor positioning to attach a popover to the button invoking it. If the popovertarget (and the anchor) are inside a sticky container, e.g. a sticky header element, the popovertarget that is positioned with anchor positioning is moving downwards instead of staying at its fixed position when the user scrolls the page.

**To Reproduce**
I created a video of it (see below) and a CodePen to isolate and reproduce the bug – or am I missing something?
https://codepen.io/matthiasott/pen/MYadqvp/d5423e2b4813be270a2b11339a0d9fe9

**Expected behavior**
The [spec](https://drafts.csswg.org/css-anchor-position-1/) says that “The [remembered scroll offset](https://www.w3.org/TR/css-anchor-position-1/#remembered-scroll-offset) also [should account] for other scroll-dependent positioning changes, such as [position: sticky](https://www.w3.org/TR/css-position-3/#propdef-position).”
I would expect the element to stay at the same position, regardless of the scroll offset.

**Screenshots**
https://github.com/user-attachments/assets/20f6702c-7418-4c33-bf6b-3b9f5746d1a6

**Browsers tested on**
Safari on MacOS (18.3.1) and iOS (18.1.1)

**Possible fix**
I temporarily fixed the behavior on my site (see the dark mode toggle in the upper right on https://matthiasott.com) by using a `hasStickyAncestor()` function that changes the strategy from `absolute` to `fixed` in the `autoUpdate` function of `applyAnchorPositions` in polyfill.ts.
`const strategy = hasStickyAncestor(wrapper) ? 'fixed' : 'absolute';`
Idk if that’s the best way to do it. Let me know if I can/should send a PR with those changes for you to review. 

Thank you so much for all the work that went into this polyfill! 🤗

jamesnw:
@matthiasott I implemented the simpler version of this, where we just consider the element's position directly, and it appears that this may be fixed. Here's my fork of your pen-
https://codepen.io/jamessw/pen/qERazWQ/7019c41759163d5f7c9f62f4da6ec15d

If you want to test it on your site, you can import the polyfill from the preview-

```js
<script type="module">
  if (!("anchorName" in document.documentElement.style)) {
    import("https://deploy-preview-413--anchor-polyfill.netlify.app/css-anchor-positioning.js");
  }
</script>
```

[...]

PR #413, opened 2026-06-15
Handle positioned elements with fixed positioning

jamesnw:
## Description
We were always using the `absolute` strategy for Floating UI, which meant that fixed position elements were not positioned correctly. This makes the strategy dynamic, based on the positioning of the positioned element. 

## Related Issue(s)
This relates to #350, and we should check if it fixes that.

## Steps to test/reproduce
Check out [position area](https://deploy-preview-413--anchor-polyfill.netlify.app/position-area.html#fixed) and [anchor](https://deploy-preview-413--anchor-polyfill.netlify.app/#fixed) examples

### DIFF
--- a/index.html
+++ b/index.html
@@ -625,7 +625,7 @@
 }</code></pre>
     </section>
 
-    <section id="fixed-target" class="demo-item">
+    <section id="fixed" class="demo-item">
       <h2>
         <a href="#fixed" aria-hidden="true">🔗</a>
         Positioning fixed position target
--- a/index.html
+++ b/index.html
@@ -48,6 +48,7 @@
     <link rel="stylesheet" href="/anchor-inside-outside.css" />
     <link rel="stylesheet" href="/import-has-import.css" />
     <link rel="stylesheet" href="/invalid-anchor.css" />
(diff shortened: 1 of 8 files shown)


index.html
```diff
@@ -623,6 +624,37 @@ <h2>
   anchor-name: --my-position-anchor-b;
 }</code></pre>
     </section>
+
+    <section id="fixed-target" class="demo-item">
```
jgerigmeyer:
```suggestion
    <section id="fixed" class="demo-item">
```

jgerigmeyer (approved):
Works for me! Interesting, in current Chrome (no polyfill) the popover does not scroll for me, but in current FF (also no polyfill) it does scroll with the page. Unrelated, but notable.

jamesnw:
> Works for me! Interesting, in current Chrome (no polyfill) the popover does not scroll for me, but in current FF (also no polyfill) it does scroll with the page. Unrelated, but notable.

I spent some time trying to isolate that bug, but wasn't able to. I think it's some combo of fixed positioning, scrolling, and `anchor()`. `position-area` seems to work.

PR #415, opened 2026-06-18
fix: support position-anchor on a custom-element host

### LINKED ISSUE #408: [BUG] `position-anchor` pointing from the `:host` of a custom element to another element is broken

jpzwarte:
Make `position-anchor` work when set on a custom-element host whose `anchor()` functions live in the shadow root's `:host` rule — including **multiple** such hosts on a page, and hosts that share a single constructed stylesheet.

## Positioning a shadow host

- Resolve `:host` / `:host(...)` target rules to the shadow host element, which `querySelectorAll` cannot return (`hostMatchesSelector`).
- Extend the anchor search to the target's own tree when a shadow host sets `position-anchor` in its own inline style, so the anchor in the outer (light DOM) tree is found — without leaking shadow-scoped names across the boundary.

## Concurrent, per-host polyfill runs

Each shadow root that adopts an anchor stylesheet schedules its own `polyfill()` run (`shadow.ts`). Those runs are async and were interleaving over shared state, so with two `<position-anchor-on-host>` hosts only the last one positioned correctly. The runs are now safe to overlap — with each other and with a user-initiated `polyfill()` — without any queue or coordinator:

- **Reentrant anchor resolution** — `parseCSS` snapshots the `anchorNames`/`anchorScopes` registries before its first `await` and passes them to `getAnchorEl`, so a concurrent run's `resetStores()` can no longer wipe the registries a run is resolving against mid-flight.
- **Stable `data-has-inline-styles` ids** — reused instead of re-minted each run (and no longer stripped on cleanup), so a concurrent run's anchor selector can't go stale.
- **Non-clobbering inline writes** — when rewriting an element's inline style, any `--anchor-*` mappings another run added are preserved (re-read at write time).

## Shared constructed stylesheets

A constructed stylesheet adopted by multiple shadow roots ("construct once, adopt everywhere") is no longer mutated in place — each adopting root receives a polyfill-owned copy with its own rewritten custom properties, so all hosts resolve, not just the last.

## Demo & tests

The `<position-anchor-on-host>` demo now uses one shared stylesheet across two linked hosts, and the e2e test asserts that **every** host is positioned above its own anchor — not just the last. Verified across many real-browser (Firefox) runs.

Closes #408

### DIFF
--- a/shadow-dom.html
+++ b/shadow-dom.html
@@ -60,6 +60,33 @@
           }
         }
         customElements.define('anchor-adopted-styles', AnchorAdoptedStyles);
+
+        class PositionAnchorOnHost extends HTMLElement {
+          connectedCallback() {
+            this.attachShadow({ mode: 'open' });
+
+            const sheet = new CSSStyleSheet();
+            sheet.replaceSync(`
+              :host {
+                --element-color: var(--target, var(--outer-anchored));
+                background: var(--element-color);
+                border: thin solid var(--border);
+                border-radius: var(--radius-1);
+                color: white;
+                font-weight: bold;
+                top: anchor(top);
+                left: anchor(center);
+                padding: 0.5em;
+                white-space: nowrap;
+                position: absolute;
+                translate: -50% -100%;
+              }
+            `);
+            this.shadowRoot.adoptedStyleSheets = [sheet];
+            this.shadowRoot.innerHTML = '<slot></slot>';
+          }
+        }
+        customElements.define('position-anchor-on-host', PositionAnchorOnHost);
       }
 
       const btn = document.getElementById('apply-polyfill');
@@ -82,7 +109,8 @@
           defineCustomElements();
 
           // Load the polyfill explicitly for the <anchor-web-component> demo.
-          // The <anchor-adopted-styles> is already covered by the adopted
+          // The <anchor-adopted-styles> and <position-anchor-on-host> demos use
+          // `adoptedStyleSheets`, so they are already covered by the adopted
           // stylesheet patch.
           await polyfill({
             roots: [document.querySelector('anchor-web-component').shadowRoot],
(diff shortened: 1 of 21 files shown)


src/dom.ts
```diff
@@ -258,11 +258,42 @@ export const getOffsetParent = async (el: HTMLElement) => {
...
+ */
+function hostMatchesSelector(host: HTMLElement, selector: string): boolean {
+  const trimmed = selector.trim();
+  if (trimmed === ':host') {
+    return true;
+  }
+  const match = /^:host\(\s*(.+?)\s*\)$/.exec(trimmed);
+  if (match) {
+    try {
+      return host.matches(match[1]);
```
jpzwarte, line 275: I reverted this line to the original one; What's in between `:host(...)` can be passed to `matches()` just fine afaik. I don't see a need to parse the CSS here (i tested the `.blue, .red` manually against `matches()`)

src/fetch.ts
```diff
@@ -89,8 +89,12 @@ function fetchInlineStyles(elements?: HTMLElement[]) {
   elementsWithInlineAnchorStyles
     .filter((el) => el instanceof HTMLElement)
     .forEach((el) => {
-      const selector = nanoid(12);
       const dataAttribute = 'data-has-inline-styles';
+      // Reuse an existing id rather than minting a new one each run: a
+      // concurrent run (e.g. another shadow root being polyfilled) may already
+      // be relying on this element's id in an anchor selector, and re-stamping
+      // it would invalidate that selector.
+      const selector = el.getAttribute(dataAttribute) ?? nanoid(12);
```
jamesnw, line 97: I think this could be useful for #91 as well.

src/parse.ts
```diff
@@ -289,12 +313,28 @@ async function getAnchorEl(
...
+  // for them.
+  let roots = options.roots;
+  const targetRoot = targetEl?.getRootNode();
+  if (
+    anchorNameFromOwnTree &&
+    (targetRoot instanceof Document || targetRoot instanceof ShadowRoot) &&
+    !roots.includes(targetRoot)
+  ) {
+    roots = [...roots, targetRoot];
+  }
```
jamesnw, line 331: Does this conceptually belong inside the `validatedForPositioning` function? I would like that function to be as close to an implementation of https://drafts.csswg.org/css-anchor-position-1/#target as possible.

src/utils.ts
```diff
@@ -115,12 +116,55 @@ export function getAdoptedStylesheetText(sheet: CSSStyleSheet) {
...
+  roots?: AnchorPositioningRoot[],
+): CSSStyleSheet {
+  if (!polyfillOwnedSheets.has(sheet)) {
+    const adopters = (roots ?? []).filter(
+      (root): root is Document | ShadowRoot =>
+        'adoptedStyleSheets' in root &&
+        (root as Document | ShadowRoot).adoptedStyleSheets.includes(sheet),
+    );
+    if (adopters.length) {
+      const copy = new CSSStyleSheet();
```
jamesnw: Are we able to maintain the [parameters](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet#parameters) from the original constructor? Claude suggests this, but I haven't verified.

```suggestion
   const copy = new CSSStyleSheet({
  media: sheet.media,
  disabled: sheet.disabled,
  baseURL: ...,   // not directly readable; media+disabled are the practical ones
});
```

jamesnw (changes requested):
This is looking good, and close! Just a few questions, thanks for your work!

shadow-dom.html
```diff
@@ -238,13 +271,13 @@ <h2>
...
+>&lt;anchor-adopted-styles&gt;&lt;/anchor-adopted-styles&gt;
+&lt;script&gt;
+&lt;!-- Load the shadow entrypoint before defining custom elements,
 so the replaceSync and adoptedStyleSheets patches are installed
 before any connectedCallback runs. --&gt;
-&lt;script type="module" src="/dist/css-anchor-positioning-shadow.js"&gt;&lt;/script&gt;
-
-&lt;anchor-adopted-styles&gt;&lt;/anchor-adopted-styles&gt;
-&lt;script&gt;
+import { patchAndPolyfillConstructedStylesheets } from '@oddbird/css-anchor-positioning/fn';
```
jamesnw, line 279: @jpzwarte I updated this - can you double check my work? Thanks!

src/parse.ts
jpzwarte: Partly. The rule it encodes — that an `anchor-name` reference resolves only within the tree scope where it was specified — is genuinely spec-#target territory. But the block decides this from `anchorNameFromOwnTree`, which is only knowable while resolving the name (inline style vs :host vs explicit anchor()). That's `getAnchorEl`'s job.

So the current split is principled:
- `getAnchorEl` resolves the reference → name + the tree scope it searches (as roots).
- `validatedForPositioning` implements #target's candidate gathering + "acceptable anchor element" validation over those roots.

Moving the block in as-is would force `validatedForPositioning` to know how the name was obtained — coupling it to name-resolution and making it less of a clean #target mirror.

shadow-dom.html
jpzwarte: LGTM. I found another code example that was out-of-date and fixed that as well.

src/utils.ts
jpzwarte: Done

jamesnw (approved):
Looks good! Thanks!

PR #423, opened 2026-06-26
Add padding to demo elements, mark non-working example

jamesnw:
## Description
Some examples were appearing different due to lack of space around a containing block, and the changes from safe alignment here- https://anchor-positioning.oddbird.net/position-area#no-block-space

I also noted an example that no longer works in polyfilled and non-polyfilled environments.

### DIFF
--- a/index.html
+++ b/index.html
@@ -717,8 +717,9 @@
       </div>
       <div class="note">
         <p>
-          With polyfill applied: Target is positioned below the Anchor, aligned
-          to its bottom right corner.
+          With polyfill applied: Target is positioned below and to the right of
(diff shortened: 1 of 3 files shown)


jgerigmeyer (approved):
> I also noted an example that no longer works in polyfilled and non-polyfilled environments.

Worth filing an issue for that separately?

jamesnw:
> > I also noted an example that no longer works in polyfilled and non-polyfilled environments.
> 
> Worth filing an issue for that separately?

It's this issue -  https://github.com/oddbird/css-anchor-positioning/issues/279, although the demo seems even more broken than initially reported.

PR #424, opened 2026-06-27
[#279] Fix position-try demos and polyfill logic

### LINKED ISSUE #279: [BUG] Position-try-fallbacks demo breaks at certain screen widths

jgerigmeyer:
Resolves #279

### Background

The `position-try` demos didn't update on scroll, and were broken in **native** browsers too, not just under the polyfill. Investigating in real native Chrome (150) confirmed two distinct problems:

1. **Demo authoring bug (affects native).** Per spec, `position-try` options are evaluated by checking whether the target overflows its **inset-modified containing block** — *not* the scrollport of an ancestor scroll container. Every demo target's containing block was its `.scroll-container` (`position: relative`), which also contains the anchor. So scrolling moved the anchor and target together and the target never overflowed its containing block → fallbacks never triggered, natively or otherwise. The polyfill only *appeared* to work because it measured overflow against the scrollport, diverging from native.

2. **Polyfill inheritance leak.** Properties shifted into custom properties (`height`, insets, margins, `anchor-name`, …) are all non-inherited in CSS, but custom properties inherit by default. So `height: 400px` on a `.scroll-container` leaked through `--height-<uuid>` into every descendant target and got stamped into every generated fallback.

### Changes

- **Demos** (`public/position-try*.css`): make the position-try scroll containers `position: static` so each target's containing block is promoted to the surrounding `.demo-elements` wrapper (an ancestor *outside* the scroller). Now scrolling pushes the target to overflow and fallbacks trigger — natively and under the polyfill. Added `overflow: clip` on the wrapper so a target whose anchor has scrolled out of view doesn't float over other content, and gave the combined demo's anchor an offset so its base position fits at rest. Removed the "this example is broken" warning in `index.html`.

- **Non-inherited custom properties** (`src/cascade.ts`, `src/dom.ts`): register the shifted custom properties with `CSS.registerProperty({ syntax: '*', inherits: false })` so a value set on an ancestor is no longer read back as if it were set directly on a descendant. Registration is global and runs once; it no-ops where `CSS.registerProperty` is unavailable. Updated the now-inaccurate inheritance comments.

- **Overflow basis** (`src/polyfill.ts`): rewrote `checkOverflow` to measure the target's margin-box against its containing block (`offsetParent`) using `getBoundingClientRect`, matching the spec/native behavior instead of floating-ui's scrollport-based `detectOverflow`. Both rects are read in viewport coordinates, so the page's scroll offset cancels out (the previous standalone `detectOverflow` call produced wildly wrong values once the containing block was no longer a near-viewport scroll container). Dropped the now-unused `detectOverflow`/`MiddlewareState` imports and `platformWithCache`.

- **Test** (`tests/e2e/polyfill.test.ts`): the target's `offsetParent` is now the (non-scrollable) wrapper, so the `@position-fallback` test scrolls `.scroll-container` directly and captures the pre-polyfill position rather than hardcoding it.

### Behavior change to flag for review

`checkOverflow` now matches native semantics (overflow vs. containing block). This preserves existing behavior for the common cases — an absolutely-positioned target whose containing block is a scroll container still flips against that scroller's visible box, and a fixed target flips against the viewport — so nothing in the suite regressed. But it is a change to core polyfill logic and worth a careful look.

### Verification

- **Native (Chrome 150):** all position-try targets now flip on scroll. The combined demo flips but non-monotonically, which is inherent to its dual-inline-axis options plus native scroll-offset compensation.
- **Polyfill (Chromium 124):** all five targets flip cleanly on scroll; no `height: 400px` leaks into generated fallbacks; no console errors.


src/fallback.ts
```diff
@@ -576,11 +576,19 @@ export function parsePositionFallbacks(styleData: StyleData[]) {
         // Parse `position-try`, `position-try-order`, and
         // `position-try-fallbacks` declarations
         const { order, options } = getPositionFallbackValues(node);
-        const anchorPosition: AnchorPosition = {};
-        if (order) {
-          anchorPosition.order = order;
-        }
```
jgerigmeyer, line 582: Claude's explanation of this change:

In the polyfill, a comma-separated selector list that uses `position-try-fallbacks` (e.g. `#a, #b { position-try-fallbacks: flip-block, … }`) caused the **second** target to be positioned using the **first** target's fallbacks. Visibly, in Firefox the second target stayed static (un-anchored) and targets could revert to a non-anchored position when a fallback couldn't be placed.

### Cause

In `parsePositionFallbacks`, the `anchorPosition` object was declared **outside** the `selectors.forEach` loop, so it was shared across every selector in the list. Because try-tactic fallbacks are keyed per selector (`${selector}-${tactics}`), each selector's fallbacks accumulated into that one shared object — and since `validPositions[selectorA]` and `validPositions[selectorB]` both referenced it, the second target's fallback list began with the first target's fallbacks (which resolve against the wrong anchor). `@position-try` (at-rule) fallbacks were unaffected because their names are shared across selectors.

### Fix

Give each selector its own `anchorPosition`, and separate the two concerns that were conflated under a single dedup guard:

- **Add a fallback to *this* selector's position** — now tracked per selector.
- **Inject the generated `@position-try` rule into the stylesheet** — still deduped once per stylesheet.

### Tests

Added a regression unit test asserting that each selector in a shared rule gets only its own try-tactic fallback. Existing unit (186) and e2e (37) suites pass.

src/cascade.ts
```diff
@@ -29,6 +29,45 @@ export const SHIFTED_PROPERTIES: Record<string, string> = [
...
+  ) {
+    return;
+  }
+  for (const customProperty of Object.values(SHIFTED_PROPERTIES)) {
+    try {
+      CSS.registerProperty({
+        name: customProperty,
+        syntax: '*',
+        inherits: false,
+      });
```
jgerigmeyer: TIL about `CSS.registerProperty` 🎉

src/cascade.ts
```diff
@@ -29,6 +29,45 @@ export const SHIFTED_PROPERTIES: Record<string, string> = [
...
+ * `height: 400px` on a scroll container) would be inherited by descendants
+ * through the shifted custom property and incorrectly read back as if it were
+ * set directly on them — leaking, for example, into every generated position
+ * fallback. See https://github.com/oddbird/css-anchor-positioning/issues/279.
+ *
+ * Registered properties are global to the document (including shadow trees), so
+ * this only needs to run once.
+ */
+let propertiesRegistered = false;
+export function registerShiftedProperties() {
```
jamesnw: Instead of eagerly registering all properties, would it make sense to check if it's registered in `shiftUnsupportedProperties` first, and only registering those?

src/cascade.ts
jamesnw: I was avoiding relying on CSS.registerProperty - it moves Safari support from Safari 10 to Safari 16.4 (Release date: 2023-03-27), and Firefox Support from Firefox 54 to Firefox 128 (Release date: 2024-07-09).

We would need to add this as a caveat. I'm not sure there's another way...

public/position-try-tactics-combined.css
```diff
@@ -1,5 +1,24 @@
...
+#position-try-tactics-combined .scroll-container {
+  position: static;
+}
+
+/* Clip the target at the wrapper (which tightly bounds the scroll container) so
+   a target whose anchor has scrolled out of view doesn't float over other
+   content. The wrapper is the target's containing block, so this does not
+   affect where fallbacks are evaluated. */
+#position-try-tactics-combined .demo-elements {
+  overflow: clip;
```
jamesnw: It seems like this shouldn't be required to have fallback work within a scrollport, but it seems to be the case. This is repeated a lot, I'm wondering if we should make it a class?

Also, the lack of padding on the elements (tight) is essential to making this pattern work- I think moving it in here would make sense.

src/polyfill.ts
```diff
@@ -1,7 +1,5 @@
 import {
   autoUpdate,
-  detectOverflow,
```
jamesnw, line 3: Nice, this was simple enough to replace. I wonder if this reduces the polyfill size, and if there are other simple-to-replace parts that would help with that?

jamesnw (commented):
Some testing notes-

In Firefox 124- Demo doesn't work. Adding padding (removing .tight from .demo-elements) makes it work, but the target goes outside of the scroll container before falling back. When it falls back, it still has the inheritance bug (CSS.registerProperty came to Firefox in 128). Fallback only happens on the y-axis.

In Firefox 128- Demo works. Adding padding makes the the targets go outside of the scroll container before falling back.

I'm not sure if this is because the custom properties are required to make it fallback correctly, or if the incorrect values are the making the fallback happen. We should figure out how crucial non-inheritance is to this change, and see if there's an alternate way of addressing that.

src/cascade.ts
jgerigmeyer: Yes, good call. It's already noted in the code comments, but I'll add a note to the Readme as well.

src/polyfill.ts
jgerigmeyer: Added #430 

## Build size comparison (gzip)

### Assets

| File | Before | After | Δ |
| --- | --: | --: | --: |
| `favicon-*.svg` | 0.83 kB | 0.83 kB | — |
| `shadow-dom.html` | 2.96 kB | 2.96 kB | — |
| `position-area.html` | 3.48 kB | 3.48 kB | — |
| `index.html` | 8.45 kB | 8.45 kB | — |
| `index-fn` (entry) | 0.11 kB | 0.11 kB | — |
| `modulepreload-polyfill` | 0.39 kB | 0.39 kB | — |
| `positionArea` | 0.40 kB | 0.40 kB | — |
| `main` | 0.94 kB | 0.95 kB | +0.01 kB (+1.1%) |
| `shadowDOM` | 1.57 kB | 1.58 kB | +0.01 kB (+0.6%) |
| `index-fn` (chunk) | 38.20 kB | 37.98 kB | -0.22 kB (-0.6%) |

### Bundles

| File | Before | After | Δ |
| --- | --: | --: | --: |
| `css-anchor-positioning.js` | 41.35 kB | 40.36 kB | -0.99 kB (-2.4%) |
| `css-anchor-positioning.umd.cjs` | 36.02 kB | 35.24 kB | -0.78 kB (-2.2%) |
| `css-anchor-positioning-fn.js` | 41.60 kB | 40.62 kB | -0.98 kB (-2.4%) |
| `css-anchor-positioning-fn.umd.cjs` | 36.33 kB | 35.53 kB | -0.80 kB (-2.2%) |

The four main bundles each shed ~2.2–2.4% gzipped, driven by the index-fn chunk shrinking.

src/cascade.ts
jgerigmeyer: Hm, in this case I think the cost of running it once initially is very low, and moving it would require a new tracking `Set` in a fn (`shiftUnsupportedProperties`) that is called multiple times. I don't think the tradeoff is worth it.

public/position-try-tactics-combined.css
jgerigmeyer: Good idea on both counts -- adjusted in https://github.com/oddbird/css-anchor-positioning/pull/424/commits/4f781ef3b4476ddb5aa642bb3adf926e388de558. You understand this better than me, so feel free to push back or adjust.

jgerigmeyer: > It seems like this shouldn't be required to have fallback work within a scrollport, but it seems to be the case.

Claude says:

> The `overflow: clip` genuinely is needed here, but only because these are _demo_ pages where a target whose anchor has scrolled out of view would otherwise float over neighboring content — it's a demo-presentation concern, not something the polyfill itself requires. The comment in `demo.css` now says exactly that.

jgerigmeyer:
> Some testing notes-
> 
> In Firefox 124- Demo doesn't work. Adding padding (removing .tight from .demo-elements) makes it work, but the target goes outside of the scroll container before falling back. When it falls back, it still has the inheritance bug (CSS.registerProperty came to Firefox in 128). Fallback only happens on the y-axis.
> 
> In Firefox 128- Demo works. Adding padding makes the the targets go outside of the scroll container before falling back.
> 
> I'm not sure if this is because the custom properties are required to make it fallback correctly, or if the incorrect values are the making the fallback happen. We should figure out how crucial non-inheritance is to this change, and see if there's an alternate way of addressing that.

@jamesnw What do you think of the fallback added in https://github.com/oddbird/css-anchor-positioning/pull/424/changes/4f781ef3b4476ddb5aa642bb3adf926e388de558..a5e9543039f4f0366ec550bef0f0d99688f3c8fc? Test again and see if this fixes what you were seeing.

src/cascade.ts
```diff
@@ -29,6 +29,72 @@ export const SHIFTED_PROPERTIES: Record<string, string> = [
...
+ *
+ * On older engines we fall back to injecting a universal rule that resets each
+ * shifted property to `initial` on every element (and the `::before`/`::after`
+ * pseudo-elements we read). For an unregistered custom property, `initial` is
+ * the guaranteed-invalid value, so any element that doesn't explicitly set the
+ * property reads back as empty instead of inheriting its ancestor's value --
+ * emulating non-inheritance. The reset has the lowest possible specificity, so
+ * real declarations still win, and it only touches our private custom
+ * properties, never the native properties that drive layout. Injected into
+ * `document.head`, this reset does not reach separate shadow trees.
```
jamesnw: Would it be sufficient to also add it to the roots passed in `options.roots`?

jamesnw (commented):
I think things are largely more similar than they were. There are a few differences still with non-polyfilled browsers, but I think that may be due to other issues.

I think before approval, I'd like to see if we can handle shadow trees easily, and verify with @mirisuzanne that resetting custom properties to `initial` won't have unintended consequences.

src/cascade.ts
```diff
@@ -29,6 +29,72 @@ export const SHIFTED_PROPERTIES: Record<string, string> = [
...
+      }
+    }
+    propertiesMadeNonInheriting = true;
+  } else if (typeof document !== 'undefined') {
+    const resets = customProperties
+      .map((customProperty) => `${customProperty}: initial;`)
+      .join('\n  ');
+    const style = document.createElement('style');
+    style.setAttribute(POLYFILLED_STYLE_ATTRIBUTE, 'true');
+    style.textContent = `*,\n::before,\n::after {\n  ${resets}\n}`;
```
mirisuzanne: This is the standard way to turn off inheritance if you can't use `@property`. The only danger I see is that we're squatting on these custom properties without any explicit name-spacing. But I think that's reasonable for a polyfill? We should just make sure it's clear that these custom properties are reserved by the polyfill, and shouldn't be set manually.

mirisuzanne (approved):

src/cascade.ts
jgerigmeyer: > We should just make sure it's clear that these custom properties are reserved by the polyfill, and shouldn't be set manually.

All of the custom properties we're setting have a UUID attached, so there shouldn't be any conflicts. I think the one current exception is the `--pa-*` properties (unrelated to this PR), but it would be easy to namespace those as well.

mirisuzanne: Well, even `--pa-` is our own namespace right? That seems fine.

jgerigmeyer: Adjusted in https://github.com/oddbird/css-anchor-positioning/pull/424/commits/0179dc10b87e3d9fa1e7d84f355799eb1565eac7 just so they're consistent with all our other custom properties.

src/cascade.ts
jgerigmeyer: Good idea! I think that's addressed in https://github.com/oddbird/css-anchor-positioning/pull/424/changes/0179dc10b87e3d9fa1e7d84f355799eb1565eac7..67ee6894750bc01337f4d1738894e66353d7b7db -- take a look.

jamesnw (approved):
This looks good to me- thanks for the fixes!

PR #428, opened 2026-07-01
Add `positionAreaContainingBlock` option

jpzwarte:
## Summary

Adds a `positionAreaContainingBlock` polyfill option that controls whether a `position-area` target is wrapped with an element approximating the containing block that `position-area` natively creates. The wrapper is correct but can interfere with author CSS that depends on the target's position in the DOM tree (direct-child/sibling combinators, the original parent's flex/grid layout, etc.). This option makes that tradeoff configurable.

- **`true`** (default): always wrap the target — matches the current behavior most closely.
- **`false`**: never wrap; the polyfill computes and applies inset values directly on the target instead. Avoids the extra element, but styles that resolve against the containing block (percentage sizes, `auto`/percentage margins, percentage padding, or `stretch`/`anchor-center` self-alignment) won't match native behavior.
- **`'auto'`**: wrap only targets whose styles resolve against the containing block (same heuristics as above); position all other targets directly.

## Implementation notes

- **`polyfill.ts`**: new public option on `AnchorPositioningPolyfillOptions` (+ normalized). The full normalized `options` object is now passed to `parseCSS` so `positionAreaContainingBlock` actually reaches the parser (previously only `roots` was forwarded, which silently forced `true`). In `applyAnchorPositions`, a `position-area` target positions via `wrapper ?? target`; unwrapped targets resolve alignment to physical inset values via the new `resolveAxisInsetValues` helper.
- **`position-area.ts`**: new `PositionAreaContainingBlock` type; `isContainingBlockDependentDeclaration` heuristic + property lists; `addPositionAreaDeclarationBlockStyles` emits the wrapped/unwrapped/both declaration sets; new `markPositionAreaTarget`, `activeTargetStyles`, and `POSITION_AREA_TARGET_ATTRIBUTE` for the unwrapped path. Wrapper insets renamed `--pa-value-*` → `--pa-wrapper-*` so a shared `'auto'` selector's target insets don't collide.
- **`parse.ts`**: collects containing-block-dependent selectors during the AST walk and computes `needsWrapper` per target in `'auto'` mode.
- **`position-area.html`**: demo page supports `?no-containing-block` (`false`) and `?auto` to exercise each mode in the browser.


src/position-area.ts
```diff
@@ -36,9 +51,110 @@ export const POSITION_AREA_CASCADE_PROPERTY = '--pa-cascade-property';
...
+  value: string,
+): boolean {
+  const prop = property.toLowerCase().trim();
+  const val = value.toLowerCase();
+  if (SIZE_PROPERTIES.includes(prop)) {
+    return (
+      val.includes('%') ||
+      val.includes('stretch') ||
+      val.includes('fill-available') ||
+      val.includes('webkit-fill-available')
```
jamesnw:
```suggestion
      val.includes('-webkit-fill-available')
```

I don't think `fill-available` is a valid value anywhere, but it does probably make sense to have `-webkit-fill-available`.

src/parse.ts
```diff
@@ -338,12 +340,35 @@ async function getAnchorEl(
...
   styleData: StyleData[],
   options: NormalizedAnchorPositioningPolyfillOptions,
 ) {
   const anchorFunctions: AnchorFunctionDeclarations = {};
   const positionAreas: PositionAreaDeclarations = {};
+  const positionAreaContainingBlock =
+    options.positionAreaContainingBlock ?? true;
+  // In `'auto'` mode, the selectors whose declarations resolve against the
+  // containing block. A `position-area` target matching any of these is wrapped.
+  const containingBlockDependentSelectors = new Set<string>();
```
jamesnw: Rather than collecting all selectors, could we only check the computed values for the elements that are using position-area? This would prevent an edge case like this-

```css
.a{ margin: 20% } 
#b{ margin: 5px; }
```
I think currently this would wrap `<div class="a" id="b" />`, even though the computed margin does not use a percent. You can use `getCSSPropertyValue()` to get the raw value, since `getComputedStyle().margin` would return a length (I think?).

<details><summary>If that doesn't work...</summary>
Instead of collecting the selectors, and later matching to the elements, could you-

1. `const containingBlockDependentElements = new WeakSet<HTMLElement>();`
2. In the walk, use `querySelectorAllRoots()` to add all the elements to that set. 
3. Then, instead of calling `matchesAnySelector()`, you can just do `containingBlockDependentElements.has(element)`. 

I think this requires less DOM queries. This still seems a bit overeager to me- there may be many many elements on a page without `position-area` that still would require a containing block. 
</details>

jamesnw (changes requested):
Overall, this looks good! Just a couple minor questions.

jgerigmeyer (approved):
The logic makes sense to me! I documented a few remaining limitations with `auto` and fixed some small bugs:

**Bug fixes**
- **Inheritance leak (issue #279):** `--pa-value-*` / `--pa-wrapper-*` inset custom properties are now registered non-inherited (new `NON_INHERITED_POSITION_AREA_PROPERTIES`, wired into `registerShiftedProperties`). Previously, an inset set on one unwrapped `position-area` target could leak through inheritance onto a descendant that is itself a `position-area` target, overriding its `auto` fallback. `--pa-value-{justify,align}-self` are left inheriting by design (wrapper → target).
- **`fit-content` false-negative in `'auto'` mode:** added `fit-content` to the containing-block-dependent sizing heuristic — it clamps to the stretch-fit (containing-block) size, so a `fit-content` target now gets a wrapper.

**API / architecture**
- Default for `positionAreaContainingBlock` now normalized in one place (`normalizePolyfillOptions`); field is required on the normalized type and the `?? true` fallback removed from `parse.ts`.
- `isPositionAreaTarget` now narrows on `targetUUID` (always present) instead of the now-optional `wrapperEl`.
- Renamed the resolved-boolean `dataForPositionAreaTarget` param to `wrap` to distinguish it from the raw tri-state option.
- Introduced `PA_INSET_SIDES` as the single source of truth for the inset sides, and exported `resolveAxisInsetValues`.

**Docs (README)**
- Documented two `'auto'`-mode limitations: containing-block dependence that appears only in a `@position-try` fallback is not detected, and the wrap decision is computed once (not re-evaluated on runtime style changes).

**Tests**
- Unit tests for `resolveAxisInsetValues` (start/end/center, null edges, negative offset for oversized targets).
- Non-inheritance test for the position-area inset custom properties.
- Unwrapped-mode inset/alignment assertions in `dataForPositionAreaTarget`; fixed the `fit-content` heuristic assertion.
- Strengthened the `'auto'` wrapped-target e2e to assert `padding-right: 50%` resolves against the wrapper (half its content width).

jamesnw (approved):
This looks good! I have a minor change I'd like to make in a follow up PR, but it's non-blocking.

PR #429, opened 2026-07-06
Remove sideEffects workaround

jgerigmeyer:
This workaround was added in #405, but is no longer needed: https://github.com/vitejs/vite/issues/22620#issuecomment-4851640360


jamesnw (approved):
🚀

PR #433, opened 2026-07-21
Fix pseudo-element anchors inside shadow roots

### LINKED ISSUE #425: document.createElement usage may break polyfill in shadow
Some addressed in #422, but further investigation is needed on the other uses.

jpzwarte:
Fixes #425 

## Summary

`createFakePseudoElement` appended its generated `<style>` (the pseudo-element's `content` rule and the `display: none` hide rule) to `document.head`, which does not apply inside a shadow root. Anchoring to a `::before`/`::after` inside a shadow root therefore measured the pseudo-element incorrectly. The style is now appended to the element's own root (shadow root, else `document.head`), mirroring #422.

## Changes

- `src/dom.ts`: scope the fake-pseudo-element `<style>` to the element's root.
- `shadow-dom.html`: new standalone demo anchoring to a `::before` inside a shadow root.
- `tests/e2e/shadow-dom.test.ts`: e2e coverage for the new case.

### DIFF
--- a/shadow-dom.html
+++ b/shadow-dom.html
@@ -274,7 +274,7 @@ customElements.define("anchor-web-component", AnchorDemo);
               }
               #shadow-pseudo-anchor::before {
                 content: 'Anchor';
-                display: block;
+                display: flex;
                 width: 120px;
                 height: 100px;
                 background: var(--anchor-color);
@@ -282,7 +282,8 @@ customElements.define("anchor-web-component", AnchorDemo);
                 color: white;
                 font-weight: bold;
                 text-align: center;
-                align-content: center;
+                align-items: center;
+                justify-content: center;
                 anchor-name: --shadow-pseudo-anchor;
               }
               #shadow-pseudo-target {
--- a/shadow-dom.html
+++ b/shadow-dom.html
@@ -33,6 +33,13 @@
         class AnchorWebComponent extends HTMLElement {}
         customElements.define('anchor-web-component', AnchorWebComponent);
 
+        // Hosts a declarative shadow root whose anchor is a `::before`
+        // pseudo-element. Like `<anchor-web-component>`, it relies on the
+        // explicit `polyfill({ roots })` call below rather than an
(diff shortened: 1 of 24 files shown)


shadow-dom.html
```diff
@@ -246,6 +256,96 @@ <h2>
...
+                content: 'Anchor';
+                display: block;
+                width: 120px;
+                height: 100px;
+                background: var(--anchor-color);
+                border-radius: var(--radius-1);
+                color: white;
+                font-weight: bold;
+                text-align: center;
+                align-content: center;
```
jamesnw: Minor, but align-content in block context is about as supported as anchor positioning, and might be misleading to have as part of this example- https://web-platform-dx.github.io/web-features-explorer/features/align-content-block/

tests/e2e/shadow-dom.test.ts
```diff
@@ -133,3 +133,28 @@ test('positions every custom-element host sharing one constructed stylesheet', a
...
+
+  const anchorBox = await anchor.boundingBox();
+  const targetBox = await target.boundingBox();
+
+  // The anchor `<span>` wraps only the block `::before`, so their boxes match.
+  // With the pseudo-element correctly hidden during measurement, the target's
+  // `anchor(bottom)` resolves within the anchor's own box, so its top stays
+  // above the anchor's bottom edge. Without the fix the un-hidden 100px
+  // `::before` pushes the target well below that edge.
+  expect(targetBox!.y).toBeLessThan(anchorBox!.y + anchorBox!.height);
```
jamesnw: Perhaps I'm misunderstanding what this is testing, but this is failing, because this is always aligned- should it be close to rather than less than?? 


```suggestion
  expect(targetBox!.y).toBeCloseTo(anchorBox!.y + anchorBox!.height, 0);
```

jamesnw (changes requested):
I think the code looks good. Minor test and example change, and then good to go!

jamesnw (approved):
Looks good to me!

PR #434, opened 2026-07-21
Feat/411 alternate approach to checking containing block props

jamesnw:
## Description
This refactors how the 'auto' mode for positionAreaContainingBlock detects whether a position-area target needs a wrapper:

**Before:** During CSS parsing, it tracked which selectors had containing-block-dependent declarations (via isContainingBlockDependentDeclaration(property, value)), then resolved those selectors to elements afterward and cross-referenced against actual position-area targets using a WeakSet. It is likely that there are many elements with these declarations on a page, and only some would be positioned with `position-area`. There was a lot of parsing and checking that was never needed.

**After:** It checks each target element directly via a new hasContainingBlockDependentDeclaration(element), which reads the element's live computed style for each relevant property (using getCSSPropertyValue from dom.ts) rather than relying on parse-time selector tracking. This adds the complexity of shifting padding properties, and I suspect each check is more expensive, but it now only checks when needed. 
 
Key changes:

1. parse.ts: Removed selector-tracking/resolution logic (containingBlockDependentSelectors, containingBlockDependentTargets WeakSet population); now just calls hasContainingBlockDependentDeclaration(targetEl) directly when needed.
2. position-area.ts: isContainingBlockDependentDeclaration → hasContainingBlockDependentDeclaration, now takes an HTMLElement and checks its computed style per property list instead of taking a raw property/value pair.
3. syntax.ts: Consolidated duplicated prop arrays, added PADDING_PROPS
4. cascade.ts: SHIFTED_PROPERTIES now also includes PADDING_PROPS, so padding declarations get duplicated to a shifted custom property (needed since the new element-based check reads live styles, including polyfill-shifted ones).

## Related Issue(s)
#411


jamesnw:
Merged into #428

PR #438, opened 2026-07-22
Move alignment props to child

jamesnw:
## Description
The wrapper's `justify-self`/`align-self` custom properties (`--pa-value-*`) were being set on the wrapper element itself alongside its inset properties (`--pa-wrapper-*`), relying on CSS inheritance to carry them down to the target child. This sets them directly on the wrapper's child instead, via a `> *` selector, removing that inheritance dependency.

1. activeWrapperStyles in position-area.ts now emits two rule blocks for a wrapper: one on the wrapper itself for insets, and a separate `[wrapper-selector] > *` block for justify-self/align-self, applied directly to the target.
2. paValueProperty/paWrapperProperty in utils.ts are no longer exported as functions; they're now precomputed into paValueProperties/paWrapperProperties Maps. This keeps the used values in sync with NON_INHERITED_POSITION_AREA_PROPERTIES.

### DIFF
--- a/position-area.html
+++ b/position-area.html
@@ -183,6 +183,20 @@
       </div>
     </section>
 
+    <section class="position-area-demo-item" id="nested-alignment">
+      <h2>
+        <a href="#nested-alignment" aria-hidden="true">🔗</a>
+        <code>alignment does not leak to descendants ✅</code>
+      </h2>
+      <div style="position: relative" class="demo-elements">
+        <div class="anchor">Anchor</div>
+        <div class="target spanleft-top" id="nested-alignment-target">
+          Target
+          <span id="nested-alignment-descendant">Nested descendant</span>
+        </div>
+      </div>
+    </section>
+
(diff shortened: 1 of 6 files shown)


position-area.html
```diff
@@ -183,6 +183,20 @@ <h2>
...
+        <code>alignment does not leak to descendants ✅</code>
+      </h2>
+      <div style="position: relative" class="demo-elements">
+        <div class="anchor">Anchor</div>
+        <div class="target spanleft-top" id="nested-alignment-target">
+          Target
+          <span id="nested-alignment-descendant">Nested descendant</span>
+        </div>
+      </div>
+    </section>
```
jgerigmeyer, line 198: I'm not tied to this, if you prefer to remove it.

jgerigmeyer (approved):
Looks good! I added some tests, and enforced types on the new maps (to guard against typos causing `var(undefined)` in the CSS).

PR #441, opened 2026-07-24
fix: missing .d.ts files in dist

### LINKED ISSUE #440: [BUG] Missing types from 0.10.0 dist
Image

According to `package.json`, the types should be in `dist/index.d.ts` and `dist/index-fn.d.ts`, but they seem to be missing from the 0.10.0 build.

jpzwarte:
Fixes #440

`package.json` points `types` and `exports` at `dist/index.d.ts` and `dist/index-fn.d.ts`, but `npm run types` was emitting them to `dist/src/` instead, so consumers got no types.

The cause is `rootDir: "."` in `tsconfig.json`. With the root as the base, `tsc` preserves the `src/` path segment in the output, turning `src/index.ts` into `dist/src/index.d.ts`.

Setting `rootDir` to `./src` puts the declarations where `package.json` expects them. `tests/tsconfig.json` extends the root config and includes files outside `src/` (tests and root-level configs), so it overrides `rootDir` back to the project root — it's `noEmit`, so the value only needs to cover its inputs.

`npm run types` now produces `dist/index.d.ts` and `dist/index-fn.d.ts`; `npm run tsc` and `npm run tsc:tests` both pass.

### DIFF
--- a/tests/tsconfig.json
+++ b/tests/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "rootDir": "..",
     "isolatedModules": false,
     "types": ["vitest/globals"],
     "noEmit": true,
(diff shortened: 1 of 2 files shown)


jamesnw (approved):
Thanks for catching this! LGTM!

PR #443, opened 2026-07-24
fix: pass polyfill options through to constructed-stylesheet shadow runs

### LINKED ISSUE #442: [BUG] `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS` is ignored when `polyfill()` is called from `patchHostConnectedCallback`
In `patchHostConnectedCallback()`, it calls the polyfill method with the `roots` option. This causes the option normalizing in polyfill to ignore `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS`. That in turn makes it impossible to use the new `positionAreaContainingBlock` option inside the shadow DOM.

Suggested solution: include `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS` when calling `polyfill` inside `patchHostConnectedCallback`.

jpzwarte:
Fixes #442

The polyfill runs set up by `patchAndPolyfillConstructedStylesheets()` called `polyfill({ roots: [shadowRoot] })` with an explicit options object, so `polyfill()` never fell back to `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS`. Global options such as `positionAreaContainingBlock: false` were silently ignored for shadow roots with adopted stylesheets — their `position-area` targets were still wrapped in `<polyfill-position-area>`.

`patchAndPolyfillConstructedStylesheets()` now accepts options and passes them down to `patchHostConnectedCallback()`, which forwards them to each `polyfill()` run, overriding `roots` with the shadow root being positioned. The options default to `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS`, matching `polyfill()`, so the no-argument case picks up the global.

The new parameter is a backward-compatible addition to the public `/fn` export.

Includes e2e tests covering both paths: options set globally, and options passed explicitly (with a conflicting global, so the explicit argument has to win).

### DIFF
--- a/src/shadow.ts
+++ b/src/shadow.ts
@@ -1,4 +1,4 @@
-import { polyfill } from './polyfill.js';
+import { type AnchorPositioningPolyfillOptions, polyfill } from './polyfill.js';
 import { captureAdoptedStylesheetText, originalReplaceSync } from './utils.js';
 
 interface CustomElementHost extends HTMLElement {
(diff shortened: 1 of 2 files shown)


jamesnw:
@jpzwarte Is there a reason to not also support passing in options to patchAndPolyfillConstructedStylesheets()? I'm trying to remember why we didn't support that when we implemented it.

jpzwarte:
> Is there a reason to not also support passing in options to patchAndPolyfillConstructedStylesheets()?

No, we could do that as well. Shall i change it to that?

jamesnw:
> > Is there a reason to not also support passing in options to patchAndPolyfillConstructedStylesheets()?
> 
> No, we could do that as well. Shall i change it to that?

Yes, please! I think parity makes sense.

jpzwarte:
> Yes, please! I think parity makes sense.

Done, please review.

jamesnw (approved):
Thanks for making this change!

PR #446, opened 2026-07-25
fix: insert position-area mapping styles into each target's own root

### LINKED ISSUE #444: [BUG] position-area in a :host rule doesn't position the host
`position-area` in a `:host` rule positions the host, which lives in the outer tree. But the generated `--pa-value-*` mapping rule is inserted into the shadow root, where it can't match the host, so the values stay undefined and the host is never positioned.

jpzwarte:
Fixes #444 

`position-area` in a `:host` rule positions the shadow host, which lives in the outer tree, but the generated `--pa-value-*` mapping rule was inserted into the roots being polyfilled. A `<style>` in a shadow root never matches that root's own host, so the values stayed undefined and the host was left unpositioned.

Record the style container of each target while the rules are generated, and insert the stylesheet into those containers instead. Targets inside a polyfilled shadow root resolve to the same container as before.

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -48,6 +48,7 @@ import {
 import {
   type DeclarationWithValue,
   generateCSS,
+  type GeneratedStyles,
   getAST,
   getRootStyleContainer,
   getSelectors,
@@ -810,15 +811,13 @@ export async function parseCSS(
     }
   }
 
-  // Create a new stylesheet for the position-area mapping styles
-  const positionAreaMappingStyleElement: StyleData = {
-    el: document.createElement('link'),
-    changed: false,
-    created: true,
+  // Collect the position-area mapping styles the polyfill generates. These are
+  // returned rather than added to `styleData`: they are polyfill output, not
+  // author styles to be rewritten in place.
+  const positionAreaStyles: GeneratedStyles = {
     css: '',
     containers: new Set(),
   };
-  styleData.push(positionAreaMappingStyleElement);
 
   // We loop through each selector that has been used to apply a position-area
   // declaration, and find all elements that match the selector. The same
@@ -864,19 +863,16 @@ export async function parseCSS(
         const activeStyles = needsWrapper
           ? activeWrapperStyles
           : activeTargetStyles;
-        positionAreaMappingStyleElement.css += activeStyles(
+        positionAreaStyles.css += activeStyles(
           targetData.targetUUID,
           positionData.selectorUUID,
         );
-        positionAreaMappingStyleElement.changed = true;
         // These rules match the target (or the wrapper inserted next to it), so
         // they belong in the target's own tree. That is not necessarily one of
         // the roots being polyfilled: a `position-area` in a `:host` rule
         // targets the shadow host, which lives outside the shadow root the
         // declaration came from.
-        positionAreaMappingStyleElement.containers?.add(
-          getRootStyleContainer(targetEl),
-        );
+        positionAreaStyles.containers.add(getRootStyleContainer(targetEl));
         // Populate new data for each anchor/target combo
         validPositions[targetSel] = {
           ...validPositions[targetSel],
@@ -893,5 +889,10 @@ export async function parseCSS(
     }
   }
 
-  return { rules: validPositions, inlineStyles, anchorScopes };
+  return {
+    rules: validPositions,
+    inlineStyles,
+    anchorScopes,
+    positionAreaStyles,
+  };
 }
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -49,6 +49,7 @@ import {
   type DeclarationWithValue,
   generateCSS,
   getAST,
+  getRootStyleContainer,
   getSelectors,
   isAnchorFunction,
   type StyleData,
@@ -815,6 +816,7 @@ export async function parseCSS(
     changed: false,
     created: true,
     css: '',
+    containers: new Set(),
   };
   styleData.push(positionAreaMappingStyleElement);
 
@@ -867,6 +869,14 @@ export async function parseCSS(
           positionData.selectorUUID,
         );
         positionAreaMappingStyleElement.changed = true;
+        // These rules match the target (or the wrapper inserted next to it), so
+        // they belong in the target's own tree. That is not necessarily one of
+        // the roots being polyfilled: a `position-area` in a `:host` rule
+        // targets the shadow host, which lives outside the shadow root the
+        // declaration came from.
+        positionAreaMappingStyleElement.containers!.add(
+          getRootStyleContainer(targetEl),
+        );
         // Populate new data for each anchor/target combo
         validPositions[targetSel] = {
           ...validPositions[targetSel],
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -814,10 +814,7 @@ export async function parseCSS(
   // Collect the position-area mapping styles the polyfill generates. These are
   // returned rather than added to `styleData`: they are polyfill output, not
   // author styles to be rewritten in place.
-  const positionAreaStyles: GeneratedStyles = {
-    css: '',
-    containers: new Set(),
-  };
+  const positionAreaStyles: GeneratedStyles = new Map();
 
   // We loop through each selector that has been used to apply a position-area
   // declaration, and find all elements that match the selector. The same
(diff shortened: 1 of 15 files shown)


tests/e2e/shadow-dom.test.ts
```diff
@@ -216,6 +216,94 @@ test('positions every custom-element host sharing one constructed stylesheet', a
...
+  // shadow root the declaration came from. The polyfill generates a stylesheet
+  // mapping the computed insets onto the target; it has to be inserted into the
+  // host's own tree, since a `<style>` inside the shadow root never matches the
+  // host. Without that, the `--pa-value-*` custom properties the target's
+  // insets read stay undefined and the host is left unpositioned.
+  await applyPolyfill(page);
+
+  // The page has already installed the adopted-stylesheet patches, so adopting
+  // a sheet into this element's shadow root queues a polyfill run for it.
+  await page.evaluate(() => {
```
jamesnw, line 233: Can you add an example to the shadow-dom.html page? It would be useful for verifying parity with non-polyfilled browsers.

jamesnw (changes requested):
Just a brief look so far. I would love a reproduction I can play with, as it isn't clear to me that this should work. I know anchor references can't cross shadow boundaries, but once an element has a default anchor, can it use it across shadow boundaries? Does that also work if you use `anchor()` without specifying an anchor?

Also, is this limited to `position-area` because we don't need to write custom styles for `anchor-size()` or `anchor()`?

jpzwarte:
> Just a brief look so far. I would love a reproduction I can play with, as it isn't clear to me that this should work.

I'm adding an example for this specific case. Without the fix you get this:

CleanShot 2026-07-27 at 12 01 19@2x

With the fix: 

CleanShot 2026-07-27 at 12 04 00@2x

tests/e2e/shadow-dom.test.ts
jpzwarte: Done

src/transform.ts
```diff
@@ -72,14 +79,19 @@ export function transformCSS(
...
-          const containers = new Set(
-            (roots?.length ? roots : [document]).map(getRootStyleContainer),
-          );
-          for (const container of containers) {
+          // This is a new stylesheet (the position-area mapping styles). A
+          // `<style>` only applies within its own tree, so it is inserted into
+          // the tree of every element its rules match, as recorded while the
+          // rules were generated. Those are not always the roots being
+          // polyfilled: a `position-area` in a `:host` rule targets the shadow
+          // host, which lives in the outer tree. Fall back to the roots when
```
jamesnw: Is this the behavior we want? If there are no containers recorded, do we still need to add styles for this?

src/utils.ts
```diff
@@ -73,6 +73,12 @@ export interface StyleData {
   // The constructed stylesheet this data came from, when the styles were
   // adopted via `adoptedStyleSheets` rather than a `<style>`/`<link>` element.
   sheet?: CSSStyleSheet;
+  // For a polyfill-created stylesheet: the containers it must be inserted into,
+  // one per tree holding an element its rules match. A `<style>` only applies
+  // within its own tree, and the elements these rules target do not always live
+  // in the roots being polyfilled — a `:host` rule styles the shadow host,
+  // which sits in the *outer* tree.
+  containers?: Set<ShadowRoot | HTMLHeadElement>;
```
jamesnw: I was curious about making this required. It seems like the only change would be that `transformCSS` currently is removing `containers` from the returned styleData. I don't think we use it after `transformCSS`, but I think I'd prefer to have `containers` required on `StyleData`, and persist it.

jamesnw (changes requested):
I think this works. I've done a bit of thinking on whether this matches the spec.

In particular- https://drafts.csswg.org/css-anchor-position/#target:~:text=Note%3A%20An,them%2E%29

> Note: An [anchor-name](https://drafts.csswg.org/css-anchor-position/#propdef-anchor-name) defined by styles in one [shadow tree](https://dom.spec.whatwg.org/#concept-shadow-tree) won’t be seen by [anchor functions](https://drafts.csswg.org/css-anchor-position/#anchor-functions) in styles in a different shadow tree, preserving encapsulation. However, elements in different shadow trees can still anchor to each other, so long as both the anchor-name and anchor function come from styles in the same tree, such as by using [::part()](https://drafts.csswg.org/css-shadow-1/#selectordef-part) to style an element inside a shadow. ([Implicit anchor elements](https://drafts.csswg.org/css-anchor-position/#implicit-anchor-element) also aren’t intrinsically limited to a single tree, but the details of that will depend on the API assigning them.)

So `:host{ bottom: anchor(--position-area-on-host top) }` or `:host{ position-anchor: --position-area-on-host }` do not work. The example here works because we are setting a `default anchor element` with the style tag, which is in the same tree as where the anchor is defined. 

If that matches your understanding, I think this makes sense, and there's just a few small changes. Thanks!

jpzwarte:
> If that matches your understanding, I think this makes sense, and there's just a few small changes. Thanks!

It does match my understanding: `:host { top: anchor(bottom) }` works because the anchor is implicit; `:host { top: anchor(--foo bottom) }` doesn't because `anchor-name: --foo` is in a different tree.

src/transform.ts
jpzwarte: Fixed

src/utils.ts
jpzwarte: Done

jamesnw (approved):
LGTM!

src/utils.ts
```diff
@@ -73,6 +73,13 @@ export interface StyleData {
   // The constructed stylesheet this data came from, when the styles were
   // adopted via `adoptedStyleSheets` rather than a `<style>`/`<link>` element.
   sheet?: CSSStyleSheet;
+  // The containers this stylesheet must be inserted into, one per tree holding
+  // an element its rules match. A `<style>` only applies within its own tree,
+  // and the elements the rules target do not always live in the roots being
+  // polyfilled — a `:host` rule styles the shadow host, which sits in the
+  // *outer* tree. Only consulted for polyfill-created stylesheets (`created`);
+  // an empty set for author styles, which are transformed in place.
+  containers: Set<ShadowRoot | HTMLHeadElement>;
```
jgerigmeyer: I think we should make this optionally undefined. In `fetch.ts` we're coercing objects to the `StyleData` type which do not have `containers` set -- the only thing we know we have is `css`, so that should remain the only required property.

```suggestion
  containers?: Set<ShadowRoot | HTMLHeadElement>;
```

src/parse.ts
```diff
@@ -815,6 +816,7 @@ export async function parseCSS(
     changed: false,
     created: true,
     css: '',
+    containers: new Set(),
```
jgerigmeyer: Thinking through it more, this approach of pushing an almost-StyleData object onto `styleData` here is sort of a smelly workaround in the first place. If we had `parseCSS` return a separate object with the new `css` and `containers` Set, I think that might be cleaner. I'll try it out and push a commit for review.

src/utils.ts
jgerigmeyer: Adjusted in https://github.com/oddbird/css-anchor-positioning/pull/446/commits/647940e770d00cac1aacce21cae360085cb7f0ce

src/parse.ts
jgerigmeyer: I tried out this approach in https://github.com/oddbird/css-anchor-positioning/pull/446/commits/705df42ec53f4dbb467cd7bfccb804884682bbbc

@jamesnw @jpzwarte What do you think?

src/transform.ts
```diff
@@ -131,3 +112,17 @@ export function transformCSS(
...
+/**
+ * Inserts styles the polyfill generated itself (the position-area mapping
+ * styles) into each container recorded while the rules were generated.
+ */
+export function insertGeneratedStyles({ css, containers }: GeneratedStyles) {
+  if (!css) return;
+  for (const container of containers) {
+    const styleEl = document.createElement('style');
+    styleEl.setAttribute(POLYFILLED_STYLE_ATTRIBUTE, 'true');
+    styleEl.textContent = css;
```
jgerigmeyer, line 125: I guess this now writes _all_ the generated CSS to _each_ container, when as far as I can tell we already know which CSS is needed by each container. Is there a good reason not to switch to a Map keyed by each container?

jgerigmeyer: Tried this approach out in https://github.com/oddbird/css-anchor-positioning/pull/446/commits/a8b2fef2acb18ab19934dd565aef1518457847d2

jgerigmeyer (approved):

jamesnw (approved):
I think these changes make sense!

Issue #369, opened 2025-11-29
[BUG] Non-HTML elements (SVG, MathML) can not be positioned


jsnkuhn:
**Describe the bug**
Single value `position-anchor` that works in Chrome (in this case `position-area: right;`) is not working in Firefox at all with the polyfill.

**To Reproduce**
https://codepen.io/jsnkuhn/pen/NPNyxrx?editors=1100

**Expected behavior**
https://drafts.csswg.org/css-anchor-position-1/#position-area-syntax

1. from my reading of the spec a single value position anchor like `position-anchor: right;` should be valid.  
2. It should compute to a 2 value syntax. In this case `right span-all`

while chrome accepts and renders this as expected. It doesn't seem to get the computed value correct. Currently computes to `position-area: right;` I don't know if i'm reading the spec wrong or if they just haven't gotten to this yet. 

The polyfill is computing to `right` incorrectly but consistently with Chrome but this isn't rendering correctly with the polyfill.

Chrome 142: 

<img width="1262" height="625" alt="Image" src="https://github.com/user-attachments/assets/aa3de5ca-d396-40c3-ac87-2521ca70fe9d" />

Firefox 145:

<img width="1441" height="680" alt="Image" src="https://github.com/user-attachments/assets/e70fdf28-76b8-45a6-8614-b3d3f982c813" />

jamesnw:
In regards to #2 above, the issue is that the polyfill doesn't support `anchor-center`, and per the [spec](https://drafts.csswg.org/css-anchor-position-1/#position-area-alignment)-

```
If all three tracks are selected, the default alignment in that axis is [anchor-center](https://drafts.csswg.org/css-anchor-position-1/#valdef-justify-self-anchor-center).
```

jamesnw:
The relevant parts of the spec-

An `anchor-name` can be used on all "elements that generate a [principal box](https://drafts.csswg.org/css-display-4/#principal-box)"

`anchor-size()` and `anchor()` can be used on any absolutely positioned box.

Since the concept of "box" is not exposed in the browser, we'll need to do some approximation. I suspect the algorithm will be something like `el istanceof HTMLElement | el instanceof MathMLElement | el.tagName === 'SVG'`.

Children of the `<math>` element do create boxes, but children of the `<svg>` element do not. 

[MathML example](https://codepen.io/jamessw/pen/ByKVRPK)
[SVG example](https://codepen.io/jamessw/pen/ZYWRKjr)

PR #447, opened 2026-07-25
fix: add `patchCSSOM`to support setting anchor positioning properties in JS

### LINKED ISSUE #445: [BUG] anchor-name and position-anchor set via CSSOM are invisible to the polyfill
Setting `anchorName` or `positionAnchor` from JavaScript does nothing in a browser without native anchor positioning. The CSSOM drops properties it doesn't know, so no declaration ends up in the style attribute.

jpzwarte:
Fixes #445

`patchCSSOM()` (opt-in, exported from `/fn`) makes every anchor positioning property the polyfill supports settable from JavaScript: `anchor-name`, `anchor-scope`, `position-anchor`, `position-area`, and the `position-try` properties. `position-visibility` is left out, as the polyfill doesn't support it. It defines each on `CSSStyleDeclaration` and stores the value in the custom property the polyfill shifts that declaration into internally — the browser can't drop a custom property, and it's already what the polyfill reads back. `setProperty()`, `getPropertyValue()` and `removeProperty()` take the dashed names as well.

`cascadeCSS` then restores those declarations to the property they were set on, before anything parses them, so a value set through the CSSOM is indistinguishable from one written in a stylesheet. That's what makes the properties beyond `anchor-name`/`position-anchor` work at all, and it keeps the CSSOM out of every parser — no special cases in `parse.ts`, `position-area.ts` or `fallback.ts`.

`'anchorName' in element.style` becomes true once it has run. `CSS.supports('anchor-name: --a')` is unaffected, and is what the README now recommends for feature detection.

Two bugs fixed along the way, both of which predate the CSSOM work:

- Inline styles inside a shadow root were never collected: `fetchInlineStyles()` only searched `document`. It now searches the polyfill roots too — both, since a shadow-scoped run still needs its host and any light-DOM anchors.
- `transformCSS()` corrupted an element's `style` attribute whenever the polyfill generated a fallback rule for that element's inline styles. It pulled the declarations back out with a fixed-length slice, which breaks as soon as anything else shares the block, writing a mangled selector into the attribute and losing the element's real styles. Generated rules now go into a stylesheet of their own. This was already reachable from plain HTML (`<div style="position-try-fallbacks: --flip">`), where the fallback silently never applied either.

Adds demos to `index.html` and `shadow-dom.html`.


src/cssom.ts
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * Values set before this is called are not picked up; call it as early as
+ * possible, alongside the other patches.
+ */
+export function patchCSSOM() {
+  if (patched || CSS.supports('anchor-name: --a')) return;
+  patched = true;
+
+  // Record which element an inline declaration belongs to.
+  for (const proto of [HTMLElement.prototype, SVGElement.prototype]) {
```
jamesnw: Is SVGElement intentional here? See #369

src/cssom.ts
```diff
@@ -0,0 +1,220 @@
...
+// property they write. A browser without native anchor positioning does not
+// know them, and the CSSOM drops what it does not know: `el.style.anchorName =
+// '--foo'` only sets a property on the style object and never produces a CSS
+// declaration, while `el.style.setProperty('anchor-name', '--foo')` is ignored
+// outright. Reading the value back still returns it, so the assignment looks
+// like it worked. The polyfill reads the `style` attribute, so without this
+// patch it never sees anchors that are wired up from JavaScript.
+const PATCHED_PROPERTIES = {
+  anchorName: 'anchor-name',
+  positionAnchor: 'position-anchor',
```
jamesnw, line 13: Why not `positionArea` or other anchor-related properties?

src/cssom.ts
```diff
@@ -0,0 +1,220 @@
...
+  if (patched || CSS.supports('anchor-name: --a')) return;
+  patched = true;
+
+  // Record which element an inline declaration belongs to.
+  for (const proto of [HTMLElement.prototype, SVGElement.prototype]) {
+    const descriptor = Object.getOwnPropertyDescriptor(proto, 'style');
+    const getStyle = descriptor?.get;
+    if (!descriptor || !getStyle) continue;
+
+    Object.defineProperty(proto, 'style', {
```
jamesnw: This seems potentially like a performance hit. Might be worth documenting?

jamesnw (changes requested):
Mostly conceptual questions at this point- 
Idea- instead of writing the unsupported values to the `style` tag, could we shift the values to custom properties? I think that would sidestep a lot of this. We could do it like the shifted properties in `cascade.ts`, so like `--anchor-name-${INSTANCE_UUID}`, and then it might just work when we need to read it.

Does this change make it work better for Lit, for instance?

This is useful for non-shadow DOM things, correct? It might be worth not tying the examples together.

Would it be possible to expose this as an option when the polyfill is run, or does it need to be applied as a separate step, so that the application can do some work before the polyfill can be run?

src/cssom.ts
jpzwarte: No, removed.

src/cssom.ts
jpzwarte: Because `anchor-name` and `position-anchor` are the dynamic parts that i'm setting from Lit. Example:

```html
<sl-button id="button">Button</sl-button>
<sl-tooltip for="button">Tooltip</sl-tooltip>
```

Which results in:

```html
<sl-button id="button" style="anchor-name: --sl-tooltip-1">Button</sl-button>
<sl-tooltip for="button" style="position-anchor: --sl-tooltip-1">Tooltip</sl-tooltip>
```

Everything else is part of `:host`.

I could add more properties of course, but i'm not sure its worth it?

src/cssom.ts
jpzwarte: Measured it: small and flat. Overriding `HTMLElement.prototype.style` means every `el.style` access on the page goes through a JS accessor, including framework code. ns per operation, native → patched:

| operation | Chromium | Firefox | WebKit |
| --- | ---: | ---: | ---: |
| read `el.style.top` | 57 → 68 | 20 → 50 | 30 → 40 |
| write `el.style.top` | 119 → 136 | 160 → 190 | 70 → 80 |
| `setProperty('top', …)` | 100 → 119 | 180 → 220 | 80 → 90 |

That's a flat ~10–30ns per access — roughly 50k accesses to add 1ms, so an animation loop touching 500 elements at 60fps costs ~0.6ms per second. And `patchCSSOM` returns early when `CSS.supports('anchor-name: --a')`, so browsers with native support never install it.

Most of it is avoidable: the native getter returns the same `CSSStyleDeclaration` per element, so the `WeakMap.set` rewrites an existing entry on every access. A symbol-keyed back-pointer instead is near-free (native-equivalent on Firefox/WebKit, ~half the overhead on Chromium), at the cost of the declaration holding a strong reference back to its element.

Documenting it in the `patchCSSOM` docblock either way.

jpzwarte:
> Would it be possible to expose this as an option when the polyfill is run, or does it need to be applied as a separate step, so that the application can do some work before the polyfill can be run?

I think it's a matter of preference. If you run `polyfill()` after HTML has rendered, then it's too late. So in that case you would need to call `patchCSSOM()` before it. Otherwise it could be an option.

jpzwarte:
> This is useful for non-shadow DOM things, correct? It might be worth not tying the examples together.

True, but where would we put an example for it? A new `cssom.html` page?

jpzwarte:
> Idea- instead of writing the unsupported values to the `style` tag, could we shift the values to custom properties? I think that would sidestep a lot of this. We could do it like the shifted properties in `cascade.ts`, so like `--anchor-name-${INSTANCE_UUID}`, and then it might just work when we need to read it.

Done!

src/cssom.ts
jpzwarte: The above is no longer relevant since it now shifts the CSSOM properties to CSS custom properties.

src/cssom.ts
jpzwarte: Perhaps including at least `position-area` as well is a good idea. I can at least think of a web component where you could specify where it is anchored. But where do you stop?

jamesnw: There are only a handful of new properties for anchor positioning, and I'd rather just support them all.

src/cssom.ts
```diff
@@ -0,0 +1,112 @@
...
+  anchorName: 'anchor-name',
+  positionAnchor: 'position-anchor',
+} as const;
+
+const PATCHED_CSS_PROPERTIES: string[] = Object.values(PATCHED_PROPERTIES);
+
+let patched = false;
+
+/**
+ * Makes `anchor-name` and `position-anchor` settable through the CSSOM, so that
```
jamesnw, line 21: Can you do an editing pass on the comments? I'm finding much of AI comments to be overly verbose to the point of being unhelpful and less clear. There's also a lot of comments that will lose context outside of this PR, with mentions like "Now, this is needed..."

src/parse.ts
```diff
@@ -107,7 +108,16 @@ export interface TryBlock {
 }
 
 function isAnchorNameDeclaration(node: CssNode): node is DeclarationWithValue {
-  return node.type === 'Declaration' && node.property === 'anchor-name';
+  // The shifted custom property counts too: `patchCSSOM` writes anchors set
+  // through the CSSOM straight into it, so they never appear as an `anchor-name`
+  // declaration. Author `anchor-name` declarations have both by this point
+  // (`cascadeCSS` appends the shifted one), so a rule now matches twice; the
```
jamesnw, line 114: This comment is going to lose context outside of this PR. I also think it would be useful to have a helper function for this check that could be re-used in `isAnchorScope`, etc.

src/parse.ts
```diff
@@ -272,9 +282,18 @@ function getAnchorFunctionData(node: CssNode, declaration: Declaration | null) {
...
+// Matches either spelling of the declaration: the property as authored, or the
+// custom property `patchCSSOM` stores a CSSOM-set value in. Built on first use,
+// as `SHIFTED_PROPERTIES` is keyed by a UUID generated at load.
+let inlinePositionAnchorRegex: RegExp | undefined;
 function hasInlinePositionAnchor(el: HTMLElement): boolean {
   const style = el.getAttribute('style');
-  return style ? /(?:^|;)\s*position-anchor\s*:/i.test(style) : false;
+  if (!style) return false;
+  inlinePositionAnchorRegex ??= new RegExp(
+    `(?:^|;)\\s*(?:position-anchor|${SHIFTED_PROPERTIES['position-anchor']})\\s*:`,
```
jamesnw, line 293: I'm curious if we actually need to match the trailing colon - could we just check if the string `position-anchor` is in the string? Is there a risk if it's a false positive?

jamesnw (changes requested):
This is looking pretty close, I think!

>  >   This is useful for non-shadow DOM things, correct? It might be worth not tying the examples together.
> 
> True, but where would we put an example for it? A new cssom.html page?

It can just be added to index.html. I know it's super long, and eventually it would be nice to have a nicer way to add demos.

src/parse.ts
jpzwarte: Reverted back to the previous one liner.

src/parse.ts
jpzwarte: `isDeclaration()` now supports an optional `property` and is used here and elsewhere.

src/cssom.ts
jpzwarte: Done

src/cssom.ts
jpzwarte: See https://github.com/oddbird/css-anchor-positioning/pull/447/changes/5b315b0f7d2ab1306c5e9a3168fb51ab8a11255d. This was more than just adding more properties to `PATCHED_PROPERTIES`:

`patchCSSOM` works by stashing whatever you assign into a private custom property, so `el.style.positionArea = 'top'` really lands in the style attribute as `--position-area-<id>: top`. Nothing else in the polyfill knows to look there — the two properties that already worked did so because two spots in the parser had been hand-taught to recognise that private spelling. 

So rather than hand-teach five more spots (which would have made every author rule match twice, since the cascade leaves both spellings behind), this commit adds one step to `cascadeCSS` that renames the private property back to the real one before anything parses it; from there the CSSOM-set value is indistinguishable from CSS the author wrote, and every existing parser handles it for free. 

The `position-try` properties also had to be added to the shifted list so their values survive being written back to the style attribute, and that in turn exposed a bug where `transformCSS` mangled that attribute whenever the polyfill generated a fallback rule for an inline style.

src/cssom.ts
```diff
@@ -0,0 +1,104 @@
...
+// Camel-case name to define on `CSSStyleDeclaration`, mapped to the CSS
+// property it writes.
+const PATCHED_PROPERTIES: Record<string, string> = Object.fromEntries(
+  CSSOM_PROPERTIES.map((property) => [
+    property.replace(/-([a-z])/g, (_, char: string) => char.toUpperCase()),
+    property,
+  ]),
+);
+
+const PATCHED_CSS_PROPERTIES: string[] = Object.values(PATCHED_PROPERTIES);
```
jamesnw: This duplicates `CSSOM_PROPERTIES`, I think?

PR #448, opened 2026-07-25
fix: inline styles were not being shifted to custom properties

### LINKED ISSUE #435: [BUG] Inline styles are not shifted
**Describe the bug**
Some declarations are "shifted" to custom properties, so that we can read raw values later on, particularly for `position-area`. For instance, `margin-left` gets shifted to something like `--margin-left-iYzMB7a9uGpUFP_AnDLsc`. 

Inline styles don't go through the same pipeline, aren't shifted, and are not available for the parts that require it.

**To Reproduce**
Add an inline style, apply the polyfill, and see that there isn't a shifted value.

Discovered by @jpzwarte as part of #428

jpzwarte:
Fixes #435 

`fetchInlineStyles()` only collected elements whose style attribute contains anchor or position-area, so an inline `margin-left: 40px` never went through `cascadeCSS()` and never got shifted into `--margin-left-<uuid>`.

The query is now built from `SHIFTED_PROPERTIES`, so inline insets, margins, sizing, padding and self-alignment are collected as well. Shifting appends the custom property rather than replacing the declaration, so the extra elements just get `--margin-left-<uuid>: 40px` alongside what they already had.

Without the fix:
CleanShot 2026-07-27 at 12 15 30@2x

With the fix: 
CleanShot 2026-07-27 at 12 16 27@2x

Native: 
CleanShot 2026-07-27 at 12 17 09@2x


src/fetch.ts
```diff
@@ -63,8 +63,26 @@ async function fetchLinkedStylesheets(
...
+// `getCSSPropertyValue` — insets, margins, sizing, padding, self-alignment,
+// `position-area` — and not just the anchor-specific ones: a target can take
+// its `position-area` from a stylesheet while setting its margin inline.
+// `anchor` is matched on its own as well, for `anchor()`/`anchor-size()` values.
+// Built on first use rather than at module evaluation: `cascade.js` and this
+// module are part of an import cycle, so `SHIFTED_PROPERTIES` is not
+// necessarily initialized yet when this module is evaluated.
+let inlineAnchorStylesQuery: string | undefined;
+function elementsWithInlineAnchorStylesQuery() {
+  inlineAnchorStylesQuery ??= [
```
jamesnw: Are there performance implications with this length of query? 

If it helps, `anchor-scope` and `anchor-name` are in SHIFTED_PROPERTIES, and duplicated by the `style*=anchor` query. Also, all the padding variants could be caught with a single wildcard, all the margin variants, etc.

src/fetch.ts
```diff
@@ -63,8 +63,26 @@ async function fetchLinkedStylesheets(
...
-const ELEMENTS_WITH_INLINE_ANCHOR_STYLES_QUERY = '[style*="anchor"]';
-const ELEMENTS_WITH_INLINE_POSITION_AREA = '[style*="position-area"]';
+// Inline styles are collected so that `cascadeCSS` can shift their declarations
+// into custom properties, like it does for the rest of the CSS. That has to
+// cover every property the polyfill later reads back through
+// `getCSSPropertyValue` — insets, margins, sizing, padding, self-alignment,
+// `position-area` — and not just the anchor-specific ones: a target can take
+// its `position-area` from a stylesheet while setting its margin inline.
+// `anchor` is matched on its own as well, for `anchor()`/`anchor-size()` values.
+// Built on first use rather than at module evaluation: `cascade.js` and this
```
jamesnw, line 86: This isn't the first time we've run into this cycle- is there a different file org that would avoid that?

jamesnw (changes requested):
Looks good, just a couple questions on this. Thanks!

src/fetch.ts
jpzwarte: > Are there performance implications with this length of query?

Yes. Engines don't bucket `[style*=…]` by attribute presence, so all 53 clauses get tested against every element in the traversal — cost tracks document size, not how many elements actually have inline styles.

Median ms per `querySelectorAll`, 20k elements / 1k styled:

| Engine | current (53) | minimized (16) | `[style]` + one regex |
| --- | ---: | ---: | ---: |
| Chromium | 14.04 | 3.94 | 0.35 |
| Firefox | 7.75 | 2.45 | 0.15 |
| WebKit | 3.25 | 0.95 | 0.15 |

(Quadrupling the styled elements to 4k barely moves the first column — 16.9ms in Chromium — confirming it's per-element-traversed.)

> …duplicated by the `style*=anchor` query. Also, all the padding variants could be caught with a single wildcard…

Agreed, and it generalizes: any term containing another term is redundant. Applied mechanically, 53 terms → 16 (`anchor`, `left`, `right`, `top`, `bottom`, `inset`, `margin`, `width`, `height`, `block-size`, `inline-size`, `justify-self`, `align-self`, `place-self`, `position-area`, `padding`).

Matched set is unchanged: a term is dropped only when a strictly shorter term is a substring of it, so every removal chain ends at a retained term that matches the same strings. Kept terms are a subset of the originals, so no false positives either. All three variants matched identical sets in every benchmark case.

src/fetch.ts
jpzwarte: Checking the real runtime graph, there are exactly two cycles:

**1. `cascade → utils → dom → cascade`** (with `utils ↔ dom` nested inside it)

This is the one that actually bites. Probing module-body evaluation shows `SHIFTED_PROPERTIES` is not yet initialized when `dom.ts`'s body runs — it only gets away with it because the read happens inside `getCSSPropertyValue` rather than at module scope. All three edges are single-use:

- `dom → cascade` exists solely for `SHIFTED_PROPERTIES` (`dom.ts:51`)
- `utils → dom` exists solely for `strategyForElement` (`utils.ts:292` → `getCSSPropertyValue`)
- `dom → utils` exists solely for `getRootStyleContainer` (`dom.ts:105`)

**2. `parse ↔ fallback`**

Only `isIdentifier` is a real value edge — `AnchorPosition`, `AnchorPositions` and `TryBlock` are types and already erase. `parse` needs `parsePositionFallbacks`; `fallback` needs those four.

Perhaps look at improving this in a new PR?

jamesnw (approved):

src/fetch.ts
```diff
@@ -63,44 +63,74 @@ async function fetchLinkedStylesheets(
...
+  if (!inlineAnchorStylesRegex) {
+    const terms = ['anchor', ...Object.keys(SHIFTED_PROPERTIES)];
+    inlineAnchorStylesRegex = new RegExp(
+      terms
+        .filter(
+          (term) =>
+            !terms.some((other) => other !== term && term.includes(other)),
+        )
+        .join('|'),
+    );
```
jgerigmeyer, line 105: Right now this false-matches declarations like `line-height: 1.5`, `float: left`, `border-top`, etc. Could we make this more efficient by anchoring the regex to declaration starts? For example, something like this:

```suggestion
    // Match at a declaration boundary, so a term appearing in a *value* does
    // not count: `float: left` and `line-height: 1.5` are not styles we read.
    inlineAnchorStylesRegex = new RegExp(`(?:^|;)\\s*(?:${terms.join('|')})`, 'i');
```

(Also added the `i` to make this case-insensitive, which I think is what we want.)

src/fetch.ts
```diff
@@ -63,44 +63,74 @@ async function fetchLinkedStylesheets(
...
+  }
+  return inlineAnchorStylesRegex.test(el.getAttribute('style') ?? '');
+}
 // Searches for all elements with inline style attributes that include `anchor`.
 // For each element found, adds a new 'data-has-inline-styles' attribute with a
 // random UUID value, and then formats the styles in the same manner as CSS from
 // style tags.
```
jgerigmeyer: This comment is stale.

jgerigmeyer (approved):
Looks good! I left one comment for a potential efficiency gain, and I think there's another related improvement we could make in this PR...

Repeat `polyfill()` runs grow the inline `style` attributes, because we keep adding them. That isn't caused by this PR, but we've now amplified it since we're matching against more declarations.

I think we could fix this by removing stale attributes that we've added on a previous run before we re-shift on a subsequent run. Something like this in `cascade.ts` (I haven't tested this):

```diff
diff --git a/src/cascade.ts b/src/cascade.ts
--- a/src/cascade.ts
+++ b/src/cascade.ts
@@ -132,12 +132,31 @@ export function registerShiftedProperties(
   }
 }
 
+/**
+ * Remove every declaration of `property` from `block`.
+ *
+ * Used before appending a generated declaration, so that re-processing a block
+ * we already wrote doesn't accumulate a duplicate each time. Inline styles are
+ * re-parsed from the `style` attribute a previous run wrote back, so this is
+ * the difference between an idempotent run and one that grows the attribute
+ * without bound. Removing rather than skipping keeps last-wins order when a
+ * block declares the same property twice.
+ */
+function dropExistingDeclarations(block: Block, property: string) {
+  block.children.forEach((child, item) => {
+    if (isDeclaration(child) && child.property === property) {
+      block.children.remove(item);
+    }
+  });
+}
+
 /**
  * Shift property declarations for properties that are not yet natively
  * supported into custom properties.
  */
 function shiftUnsupportedProperties(node: CssNode, block?: Block) {
   if (isDeclaration(node) && SHIFTED_PROPERTIES[node.property] && block) {
+    dropExistingDeclarations(block, SHIFTED_PROPERTIES[node.property]);
     block.children.appendData({
       ...node,
       property: SHIFTED_PROPERTIES[node.property],
@@ -162,6 +181,7 @@ function expandInsetShorthands(node: CssNode, block?: Block) {
 
   const appendProperty = (property: string, value?: CssNode) => {
     if (!value) return;
+    dropExistingDeclarations(block, property);
     block.children.appendData({
       ...node,
       property,
```

Or @jpzwarte and @jamesnw if you think this deserves its own issue and PR, that's fine too!

src/fetch.ts
jgerigmeyer: This regex also probably deserves its own unit test.

jgerigmeyer: Oh wait, I just realized this suggestion doesn't fit with the existing term-reduction. We could adjust the reducer to use `startsWith`, but I think I'd recommend removing it entirely. Edited my code suggestion accordingly.

jamesnw: I ran a benchmark, and there wasn't a significant difference between a regex with 50 terms and a regex with 13. I think the real win was moving it to a regex rather than multiple `[style*=…]` selectors. So I removed the reducer, and added tests.

jamesnw:
> Repeat `polyfill()` runs grow the inline `style` attributes, because we keep adding them. That isn't caused by this PR, but we've now amplified it since we're matching against more declarations.

> Or @jpzwarte and @jamesnw if you think this deserves its own issue and PR, that's fine too!

I have some questions (do we drop all, or just overwritten, for example) so I opened #461

src/fetch.ts
```diff
@@ -63,44 +63,75 @@ async function fetchLinkedStylesheets(
...
+// do not bucket attribute-substring selectors by attribute presence, so a
+// ~50-clause query runs every substring test against every element in the
+// document; querying `[style]` and filtering here is an order of magnitude
+// faster, and scales with the number of styled elements rather than with the
+// size of the document.
+//
+// A term that contains another term is redundant -- `margin` already matches
+// `margin-inline-start`, `anchor` already matches `anchor-name` -- so only the
+// shortest distinct ones are kept.
+//
```
jgerigmeyer:
```suggestion
```

tests/unit/fetch.test.ts
```diff
@@ -246,3 +246,80 @@ describe('fetch styles manually', () => {
...
+    ['text-align: right (contains "right")', 'text-align: right;'],
+    ['outline-width (contains "width")', 'outline-width: 1px;'],
+    [
+      'background-position: top (contains "top")',
+      'background-position: top right;',
+    ],
+    ['column-width (contains "width")', 'column-width: 100px;'],
+    ['transform-origin: top left', 'transform-origin: top left;'],
+    ['term as custom property', '--anchor: anchor(--my-anchor);'],
+  ])('returns true (known false positive) for %s', (_name, style) => {
```
jgerigmeyer:
```suggestion
  ])('returns false for %s', (_name, style) => {
```

jgerigmeyer (approved):

PR #450, opened 2026-07-31
Lint example scripts

Guflly:
Adds `eslint-plugin-html` for the remaining inline scripts and moves the demo module scripts into checked `.ts` files. `tsc:demo` now catches API type mismatches in the demos.

Fixes the DOM lookup and attribute errors surfaced by the new checks.

Tests: type checks, ESLint, Prettier, unit tests, CSS lint, and all builds.

Closes #386.


jamesnw:
@Guflly Thanks for your contribution. Can you confirm it would catch the type error in the linked issue? 

Also, are you a user of the polyfill?

Guflly:
Good catch. The first version wouldn't catch that type mismatch. I moved the module scripts into checked `.ts` files and added `tsc:demo` to the lint run. I also verified that the ShadowRoot mismatch from #353 now fails the demo type check.

PR #459, opened 2026-08-12
fix: constructed-stylesheet shadow runs and missing type declarations

### LINKED ISSUE #449: [BUG] `window. ANCHOR_POSITIONING_POLYFILL_OPTIONS` is not typed
`window. ANCHOR_POSITIONING_POLYFILL_OPTIONS` is not typed, meaning TypeScript will complain about it if you use it in your application. Workaround of `(window as any).ANCHOR_POSITIONING_POLYFILL_OPTIONS` is bad since you will lose any safety checks TypeScript does for you.

Should we ship typing for `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS`?

jgerigmeyer:
## Summary

Three fixes to the constructed-stylesheets shadow-DOM path and to what ships in `dist/`.

**1. Apply polyfill options to constructed-stylesheet shadow runs** (follow-up to #443)

PR #443 fixed the symptom of #442 but left three defects behind:

- The re-patch guard compared `descriptor.set` against a value read from that same descriptor, so it was always true. Every call to `patchAndPolyfillConstructedStylesheets()` stacked another wrapper on the `adoptedStyleSheets` setter, and combined with the `patchedHosts` short-circuit, a second call's options were silently discarded.
- Options were captured when the patch was installed, so a global set afterwards was ignored — despite the docs telling callers to install the patch as early as possible.
- A global `elements` list was forwarded into the shadow runs, where it makes `fetchCSS` skip adopted stylesheets entirely, disabling the feature this function exists to provide.

Options are now read at run time; `elements` and `roots` are overridden and omitted from the accepted options type (`ConstructedStylesheetsPolyfillOptions`), so passing either is a compile error rather than a silent no-op.

**2. Position hosts that adopt a stylesheet before being connected**

`patchHostConnectedCallback()` wrapped the host element's own `connectedCallback` property. That never ran: lifecycle callbacks are looked up when the element is defined and stored on the definition, so assigning `host.connectedCallback` afterwards has no effect on the reaction. Everything kept working only via the `host.isConnected` branch, meaning a host that adopts while disconnected — the familiar pattern of building the shadow root in the constructor — was **never polyfilled at all**.

`customElements.define` is now patched to wrap `connectedCallback` on the constructor's prototype before the registry captures it. That fires wherever the host is connected, including inside another component's shadow root, which a document-level `MutationObserver` cannot see — mutation records don't cross shadow boundaries.

The trade-off, now documented in the README: a host adopting while disconnected is positioned only if it is a custom element defined after this call. The documented requirement was already to call this before defining elements; a non-custom-element host with a detached shadow root is not covered, since nothing signals that it has been connected.

**3. Ship type declarations and stop the demo build deleting them** — Fixes #449

`src/@types/global.d.ts` was an input `.d.ts`, which `tsc` does not copy to `outDir`, so consumers following the documented config example got `TS2339` on `window.ANCHOR_POSITIONING_POLYFILL_OPTIONS`. Moved to a compiled module pulled in by each entry with a side-effect import, which `tsc` preserves in declaration output.

Separately, only the library builds set `emptyOutDir: false`, so `build:demo` — the first step of `build` — wiped any declarations `npm run types` had emitted. `prepack` was correct only by ordering luck. Builds are now additive, with a `clean` step up front.

## Related

- Fixes #449
- Follow-up to #443 / #442
- Hardens #441 / #440 — `build:demo` could still delete the declarations #441 arranged to emit


jamesnw:
@jpzwarte Can you take a look at this PR? Thanks!

README.md
```diff
@@ -88,6 +88,26 @@ stylesheet source text, and patches the `ShadowRoot.prototype.adoptedStyleSheets
 setter to automatically run the polyfill for each shadow root once its host
 element's `connectedCallback` finishes.
 
+A host that adopts its stylesheet before it is connected — typically one that
+builds its shadow root in the constructor — is positioned by its
+`connectedCallback`. That requires `customElements.define()` to be patched
+first, which is why this must be called before those elements are defined. Such
+a host is also never positioned if it isn't a custom element, since nothing
+signals that it has been connected.
```
jamesnw: I think this comment could be cleaned up a bit, maybe just:

```suggestion
The polyfill only positions custom elements defined with `customElements.define()` after `patchAndPolyfillConstructedStylesheets()` is called.
```

src/shadow.ts
```diff
@@ -1,47 +1,118 @@
...
-  const host = shadowRoot.host as CustomElementHost;
-  if (patchedHosts.has(host)) {
+function patchCustomElementsDefine() {
+  if (customElementsPatched || typeof customElements === 'undefined') {
     return;
   }
-  patchedHosts.add(host);
+  customElementsPatched = true;
+
+  const originalDefine = customElements.define;
```
jpzwarte: This only works for the global custom elements registry, not any scoped custom element registries. Let's look at that in a new PR.

jpzwarte: Can we patch `CustomElementRegistry.prototype.define` instead perhaps? That would then work for scoped custom element registries as well?

jgerigmeyer: @jpzwarte Great catch! Can you test the fix pushed in https://github.com/oddbird/css-anchor-positioning/pull/459/commits/1b25a5d907464a870a47f7d6f67ef33e9366a35f?

jpzwarte: I can't contribute directly to your branch; i have to still use a fork etc. 

On my fork i've added an example to `shadow-dom.html`:

CleanShot 2026-08-16 at 11 16 37@2x
CleanShot 2026-08-16 at 11 16 45@2x

jpzwarte: I've created a PR for adding this example to your branch, but i see your branch isn't up-to-date with `main`, so the PR includes a bunch of unwanted changes. Perhaps merge `main` into this one first?

jpzwarte: To be clear: 
- First screenshot is Chrome 153: supports anchor positioning *and* scoped registries
- Second screenshot is Firefox dev edition 154: supports anchor positioning but not scoped registries

But on `main` the polyfill doesn't try and patch `CustomElementRegistry.prototype.define`, so you shouldn't se any difference there.

On this branch, you *should* see a difference if you switch between patching `CustomElementRegistry.prototype.define` and `window.customElements.define`.

jpzwarte: Now i am confused: i don't think this is testable. There aren't any browsers that *don't* support CSS Anchor Positioning yet, but *do* support Scoped Custom Element Registries. I don't think that's a real world usecase:

https://caniuse.com/css-anchor-positioning vs https://caniuse.com/wf-scoped-custom-element-registries

It's still probably better to patch `CustomElementRegistry.prototype.define`, but testing the polyfill in that combination is not possible?!

jpzwarte (approved):

src/shadow.ts
jgerigmeyer: That's a great point -- I think this branch is still worth merging, but I'm not sure #463 really adds anything useful.

jamesnw (approved):

jamesnw (approved):
This looks good to me!

jamesnw (approved):
This looks good to me!

